PR #2, opened 2023-02-28
Revert "Done until 6th step of firebase bootcamp."

brygotti:
Revert the latest merge to main, since we want to make sure main branch only contains the first week bootcamp code when we start working on tasks for the first sprint.


Thymian18 (commented):
LGTM

PR #4, opened 2023-03-02
Maps bootcamp

Thymian18:
For the google maps part, I implemented a map with a marker at the position of EPFL's Satellite. When clicking the marker, an info window pops up that is itself clickable. When clicking the info window, a toast message lets the user know the coordinates of Sat. Also, as soon as the user clicks 5 times or more on the marker, a different toast message pops up. 
In general the map starts over the EPFL area and is constrained to certain zoom factors (max and min).
Testing the map was not well feasible because the internal objects of the map (e.g., the marker) could not be accessed. This is why the coverage is below 80%.


Luca-Engel (commented):
Please remove commented out lines in the GreetingActivity class

app/src/main/res/layout/activity_greeting.xml
```diff
@@ -14,6 +14,20 @@
...
+        app:layout_constraintTop_toTopOf="parent"
+        app:layout_constraintVertical_bias="0.353" />
+
+    <Button
+        android:id="@+id/open_maps_btn"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_marginTop="100dp"
+        android:fontFamily="sans-serif-medium"
+        android:text="Google Maps"
```
Luca-Engel, line 28: use @string resource for "Google Maps

app/src/main/res/layout/activity_maps.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<fragment xmlns:android="http://schemas.android.com/apk/res/android"
```
Luca-Engel, line 2: cirrus suggests: Replace the <fragment> tag with FragmentContainerView.

app/src/test/java/com/github/sdpcoachme/TestMapsActivity.java
```diff
@@ -0,0 +1,21 @@
...
+import android.app.Activity;
+import android.os.Bundle;
+
+import com.google.android.gms.maps.GoogleMap;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class TestMapsActivity {
+    /*@Test
```
Luca-Engel, line 13: The code would be a bit more legible if the commented test was removed

Luca-Engel (approved):
LGTM, thanks for the improvements!

PR #5, opened 2023-03-02
Google Authentication bootcamp

Luca-Engel:
Google Authentication added as an activity that is reached by clicking on the "Go to sign in page" button in the main activity.
Sign in, Sign out and delete account are done by clicking on the buttons in the new activity.

Dependency injection was used to provide an example for future possible uses of it and to abstract the login interface.

Testing:
2 Tests are commented out because they only worked locally and not on the CI. I did not manage to find and fix the reason for that.


Thymian18 (approved):
LGTM

PR #6, opened 2023-03-02
Web api

Jeremmmyyyyy:
The web api adds a new activity to the app where it is possible to make requests to a webserver and cache the response. In the case where there is no internet connection a cached answer is retrieved and displayed. 

Problems : I wasn't able to completely test the code because the dependency injection was to difficult to manage / change in the test classes. 

Used tools : retrofit (api calls) / hilt (dependency injecction) / dagger (dependency injecction / room (local database)


app/src/main/java/com/github/sdpcoachme/repositories/DefaultRepositoryAccess.kt
```diff
@@ -0,0 +1,77 @@
...
+            val randElem = entriesDB[rand]
+            "No internet connection : seeing cached data : " + randElem.activity
+        }else{
+            "No internet connection : no cached data"
+        }
+        str
+    }
+
+    override fun deleteContentDB() {
+        CoroutineScope(Dispatchers.IO).launch {
```
brygotti, line 44: It might be a good idea to expose all the methods in the same format, either make them all launch a coroutine, or annotate them all with "suspend" keyword to enforce that they are called from within a coroutine.

brygotti: Another idea could be to use Futures instead and make every function return a Future<SomeType>

app/src/main/java/com/github/sdpcoachme/repositories/RepositoryAccess.kt
```diff
@@ -0,0 +1,18 @@
+package com.github.sdpcoachme.repositories
+
+import com.github.sdpcoachme.network.DataFormat
+
+import retrofit2.Response
+
+interface RepositoryAccess {
+
+    suspend fun getContentDB() : String
```
brygotti, line 9: As both what you call DB and what you call API are actually databases (the first is the "cache", and the second is the remote database), it might be better to rename them to something more distinct (I suggest cache for the local cache and remoteDB for the remote database, or something along those lines)

app/src/main/java/com/github/sdpcoachme/BoredActivity.kt
```diff
@@ -0,0 +1,66 @@
...
+        }
+    }
+
+    @SuppressLint("SetTextI18n")
+    fun request(txt : TextView){
+        CoroutineScope(Dispatchers.IO).launch {
+            val resp : Resource<Response<DataFormat>> = repo.apiCall()
+            if (resp.status == Status.SUCCESS) {
+                txt.text = "Activity : " + (resp.data?.body()?.activity ?: "Null")
+                resp.data?.let { it1 -> repo.insertNewEntryDB(it1) }
```
brygotti, line 57: It might be better to hide the code that saves the data in the cache from the main activity, and make it private to the repository class (the cache is usually transparent to the user of the class)

app/src/main/java/com/github/sdpcoachme/database/RequestPoolDB.kt
```diff
@@ -0,0 +1,17 @@
...
+
+@Dao
+interface RequestPoolDB {
+    @Query("select * from LineDB")
+    fun getAll(): List<LineDB>
+
+    @Insert
+    fun insert(vararg lineDB: LineDB)
+
+    @Query("DELETE FROM lineDB")
```
brygotti, line 15: Difference between lineDB and LineDB ?

app/src/main/java/com/github/sdpcoachme/repositories/Resource.kt
```diff
@@ -0,0 +1,17 @@
+package com.github.sdpcoachme.repositories
+
+data class Resource<out T>(val status : Status, val data : T?, val message: String?){
```
brygotti, line 3: Seems like this class could be replaced by using built-in Futures

app/src/main/java/com/github/sdpcoachme/network/DataFormat.kt
```diff
@@ -0,0 +1,11 @@
+package com.github.sdpcoachme.network
+
+data class DataFormat(
```
brygotti, line 3: Redundancy between this and LineDB

brygotti (changes requested):
If we decide to merge this, we will need to take care of dependency injection afterwards. Since it currently doesn't work, I didn't review the files linked to it. I let the request pending so we don't forget to refactor the different changes suggested in the comments before merging.

Ayskin:
Close for now. Reopen if we need any web api that cannot be integrated within android studio

PR #20, opened 2023-03-04
Add code to enable dependency injection during testing

brygotti:
This pull request doesn't add much code except the boilerplate code necessary to do dependency injection in testing.

The last commit (4213ef7a8611dc3e0f70042924c6a4514ac9d6d3) contains only the code that enables dependency injection, while commit 73da02ec25f300fcddcad27c557078ab367b689c consists of an example on how to use this code to do a DI. The example contains a rudimentary database interface, and a real and a mock implementation. One needs to take a look at 73da02ec25f300fcddcad27c557078ab367b689c to understand how to do DI in testing.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -15,7 +15,7 @@ android {
         versionCode 1
         versionName "1.0"
 
-        testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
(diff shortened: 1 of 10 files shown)


Luca-Engel (approved):
LGTM

PR #23, opened 2023-03-13
Google auth startup activity

Luca-Engel:
At start-up, ¨the user has three possibilities: sign in, sign out, or delete. All the buttons then call google authentication. 

If sign-in is called, then the method 'private fun launchPostLoginActivity(email : String)' in MainActivity.kt is called with the user's email passed as an argument.

Test coverage is not that high because I was not able to test all the google authentication parts (e.g., sign in)

MockGoogleAuthenticator theoretically not necessary but once we add a check in MainActivity to see if user already is logged in and then skip the manual sign in part, it will be useful for testing


app/src/main/java/com/github/sdpcoachme/firebase/database/RealDatabase.kt
```diff
@@ -0,0 +1,8 @@
+package com.github.sdpcoachme.firebase.database
+
+class RealDatabase : Database {
+
+    override fun isExistingUser(email: String): Boolean {
+        return true
+    }
```
Ayskin, line 7: Maybe we can add a TODO here if we need to implement this function later ?

app/src/androidTest/java/com/github/sdpcoachme/MainActivityTest.kt
```diff
@@ -1,34 +1,136 @@
...
+//        Intents.init()
+//
+//        val name = "John Lennon"
+//        composeTestRule.onNodeWithTag("textfield").performTextInput(name)
+//        Espresso.closeSoftKeyboard()
+//        composeTestRule.onNodeWithTag("button").performClick()
+//
+//        intended(allOf(hasComponent(GreetingActivity::class.java.name), hasExtra("name", name)))
+//        Intents.release()
+//    }
```
Ayskin, line 33: I think we can delete the commented correctIntentFiredOnDisplayMessageClick() test function !

Ayskin (commented):
Seems all good and clean to me !

app/src/main/java/com/github/sdpcoachme/firebase/database/RealDatabase.kt
lindsaybordier: I agree with that

app/src/main/java/com/github/sdpcoachme/MainActivity.kt
```diff
@@ -4,53 +4,112 @@ import android.content.Intent
...
+            res,
+            { email -> launchPostLoginActivity(email ?: "") },
+            { errorMsg -> signInInfo = errorMsg.toString() }
+        )
+    }
+
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
+        authenticator = (application as CoachMeApplication).authenticator
+        this.database =  (application as CoachMeApplication).database
```
lindsaybordier, line 44: This is great a very clean use of dependency injection and will merge easily in the project as it was coded the same way in the branch implementing the database.

app/src/androidTest/java/com/github/sdpcoachme/firebase/auth/GoogleAuthenticatorTest.kt
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    @Test
+    fun isSignedInReturnsFalseIfUserNotSignedIn() {
+        FirebaseAuth.getInstance().signOut()
+        assertThat(GoogleAuthenticator().isSignedIn(), `is`(false))
+    }
+
+    @Test
+    fun signIt() {
```
lindsaybordier: This test seems empty so we could maybe remove it

PR #24, opened 2023-03-14
Firebase kotlin

lindsaybordier:
Introduces the first methods to access the database, as well as buttons on the main activity to set or retrieve a value from the database on the app.

Interface Database was created along with 2 implementations, one for the real Firebase database, and a Mock database for testing.


app/src/main/java/com/github/sdpcoachme/MainActivity.kt
```diff
@@ -17,40 +18,107 @@ import androidx.compose.ui.platform.LocalContext
...
+                    horizontalAlignment = Alignment.CenterHorizontally
+                ) {
+                    FirebaseForm()
+                    GreetingForm()
+                }
+            }
+        }
+    }
+
+    //TODO put test tags in separate class
```
Thymian18, line 44: Same thing as before, if resolved remove, otherwise okay

app/src/main/java/com/github/sdpcoachme/MainActivity.kt
```diff
@@ -17,40 +18,107 @@ import androidx.compose.ui.platform.LocalContext
...
+
+
+    /**
+     * A form for testing the database
+     */
+    @Composable
+    fun FirebaseForm() {
+        var phoneText by remember { mutableStateOf("") }
+        var emailText by remember { mutableStateOf("") }
+        Column( //modifier = Modifier.fillMaxSize(),
```
Thymian18, line 54: You might want to remove those commented modifiers if the design works well without them. Otherwise, maybe a small describtion about when to uncomment those lines would be useful.

Thymian18 (changes requested):
Check the comments, you might want to remove some of them. Otherwise, very clear code.

Luca-Engel (approved):
LGTM, well coded and nicely structured!

Thymian18 (approved):

PR #25, opened 2023-03-15
Edit profile

Thymian18:
This is the front-end of the activity where a user can see his personal data and modify it. For now, there are the fields first-/last name, email, favorite sport, and profile picture, but this can be extended quite easily.


app/src/androidTest/java/com/github/sdpcoachme/EditProfileActivityTest.kt
```diff
@@ -0,0 +1,90 @@
...
+        composeTestRule.onNodeWithTag("editable fav sport")
+            .performTextInput("Updated fav sport")
+        Espresso.closeSoftKeyboard()
+
+        //save updated profile
+        composeTestRule.onNodeWithTag("save button")
+            .assertIsDisplayed()
+            .performClick()
+
+        correctInitialScreenContent()
```
brygotti, line 88: It might make sense to test that clicking the save button actually changes the text displayed for each field, that is that each field contains "Update first name", "Updated last name", ... respectively.

app/src/androidTest/java/com/github/sdpcoachme/EditProfileActivityTest.kt
brygotti, line 1: We need to modularize our testing classes. The tags need to be stored in a structured manner so that we can loop over all tags of a given activity, or only certain tags, etc. so that we don't have to do such code repetition. This doesn't need to be done before merging this PR, but an issue could be open so that we don't forget to do this. Because this testing code seems not particularly modular

app/src/main/java/com/github/sdpcoachme/EditProfileActivity.kt
brygotti, line 1: Since each row is exactly the same (a textfield with a label), it might make sense to modularize this into one Composable with all the common code, and instantiate it with different labels

brygotti (changes requested):
If possible, it might be a good idea to already modularize the UI (and maybe add the suggested test) before merging.

app/src/main/java/com/github/sdpcoachme/EditProfileActivity.kt
brygotti: One last thing, it might be enough to use a mutable state of String (instead of TextFieldValue). I don't know if there are any particular reason why you chose to go with TextFieldValue, but I think using a String is very common and usually works without the need to call .text whenever you want to access the string.

brygotti (approved):

PR #26, opened 2023-03-15
Dashboard

Ayskin:
Created a dashboard implemented via a left-side navigation drawer. Each button in the dashboard menu should lead to a specific app functionality/activity. 

TODOs are left in the code for remaining implementations, mainly the links to other activities in the menu. Each new implementation should be tested accordingly, some of these test signatures are already written.

The CI coverage may be low for yet undiscovered reason, in which case it needs to be further investigated.


Luca-Engel (approved):
LGTM

PR #28, opened 2023-03-16
Enter account details

lindsaybordier:
The functionality allowing a user to enter his account details to create an account was added.

The database was structured according to the team's agreement on how user informations and other data should be stored.
All the database methods (interface, firebase implementation, mockdatabase for testing) were implemented carefully for easy creation of more complex database access methods that will be needed in the future.

Also, a separate activity with a Multiselect List was created for easy selection of sport. The two activities will be linked together in another task this sprint.

Finally: In some further task, some checks should be implemented regarding the data entered, and notifications should appear when the data is stored i.e. when the future completes :)


Thymian18 (approved):
LGTM, all the necessary code is there and commented. In a next sprint we should add the functionality to being able to change attributes of a given user.

PR #29, opened 2023-03-16
Edit profile backend

Luca-Engel:
Enabled Writing displaying the current information of the user from the database and allowing for the user to change the information.

Also, adapted the getUser(email: String) method header since only the email is needed to identify a user and nothing else

Adapted DashboardActivity to open EditProfileActivity when the "Profile" button is pressed


app/src/androidTest/java/com/github/sdpcoachme/EditProfileActivityTest.kt
brygotti, line 1: Consider adding a test that tests whether the edit button correctly propagates the update to the database (using the mockdatabase)

app/src/androidTest/java/com/github/sdpcoachme/firebase/database/UserInfoTest.kt
brygotti, line 1: There might be a way to do this directly using the database interface! Would require much less code. We will dig into that for next sprint.

app/src/main/java/com/github/sdpcoachme/EditProfileActivity.kt
```diff
@@ -184,12 +211,14 @@ fun ProfileRow(rowName: String, isEditing: Boolean, leftTextPadding: Dp, value:
                 onValueChange = { newValue -> onValueChange(newValue) },
                 singleLine = true,
                 maxLines = 1)
+            println("the thing is: <<editable $lowercaseRowName>>")
```
brygotti, line 214: This debug code can be removed, I suppose ?

app/src/main/java/com/github/sdpcoachme/EditProfileActivity.kt
```diff
@@ -44,12 +58,23 @@ class EditProfileActivity : ComponentActivity() {
...
-    var fname by remember { mutableStateOf("Damian") }
-    var lname by remember { mutableStateOf("Kopp") }
-    var favsport by remember { mutableStateOf("Jogging") }
+    var fname by remember { mutableStateOf("") }
+    var lname by remember { mutableStateOf("") }
+    var favsport by remember { mutableStateOf("") }
+
+    var f by remember { mutableStateOf(future)}
+
+    f.thenAccept { newUser ->
```
brygotti, line 70: I am not sure, but it seems that doing future.thenAccept { ... } directly, without creating another future as a mutable state might be easier ?

app/src/main/java/com/github/sdpcoachme/EditProfileActivity.kt
```diff
@@ -44,12 +58,23 @@ class EditProfileActivity : ComponentActivity() {
  * Composable used to display the user's profile.
  */
 @Composable
-fun Profile() {
+fun Profile(email: String, database: Database, future: CompletableFuture<UserInfo>) {
```
brygotti, line 61: database doesn't need to be passed as a parameter, it can be retrieved by doing
```val context = LocalContext.current``` in the composable and then ```(context.applicationContext as CoachMeApplication).database``` to get the instance

brygotti (approved):
Since we don't have a lot of time before the end of the sprint, we can merge this and we'll refactor / add more tests later. I'll add issues linking to the review comments so we don't forget about them.

PR #32, opened 2023-03-17
Connect all activities together

brygotti:
This branch contains code that connects all activities together with intents. It is also a first refactoring of the directory structure of the project. TODOs were added where some things were left unfinished for now. The code in here will serve as the demo for sprint 2.


brygotti:
Apparently, we have less than 80% test coverage. This is due to parts of the code like the login activity being very hard to test. Such issue with test coverage should disappear in the long run as more code will be added and tested.

Ayskin:
LGTM! Agree with the use of TODOs!

Ayskin (approved):
LGTM! Agree with all the TODOs

PR #53, opened 2023-03-22
On create error handling

Luca-Engel:
Feature: Recovering from errors coming from extras missing in the Intents

The activity IntentExtrasErrorActivity has been added to the project. This activity can now be called whenever an onCreate activity that requires an extra field to be passed (e.g., the email address as a string) is launched without that extra being passed. This is done to recover from errors.
- This activity either displays a generic error or the error passed to it in the extra field of the intent that launched it. (Example image shown below)
- Then, the user can click on the button displayed on the screen to return to the LoginActivity and retry what they did before.
- This activity provides a simple way to recover from errors without crashing while explaining to the user what went wrong.
- Currently, the error activity is launched if DashboardActivity, SignupActivity, or EditProfileActivity are launched without an email passed in the intent

Testing:
- Some existing tests had to be adjusted to include an email addresse in the intents launching the activities so that the tests could test the actual activities and the error page wouldn't be launched instead
- Tests for IntentExtrasErrorActivity have also been added

Example error:
image

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -43,7 +43,7 @@ android {
         compose true
     }
     composeOptions {
-        kotlinCompilerExtensionVersion '1.2.0'
+        kotlinCompilerExtensionVersion '1.4.3'
     }
     packagingOptions {
         resources {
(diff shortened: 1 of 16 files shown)


Thymian18 (approved):
Good work, the new classes and activities have clear abstraction. The testing seems to be robust too.

PR #54, opened 2023-03-23
Use local version of jacoco to avoid issues with branch coverage and Jetpack Compose

brygotti:
This PR adds a local maven repository to the project, in the libs directory. It contains a custom version of Jacoco that fixes issues with Jetpack Compose (very low branch coverage due to "invisible" branches being detected in Composables). This Jacoco version was built locally from PR https://github.com/jacoco/jacoco/pull/1259 , and the project is configured to use it instead of the default Jacoco version from the global maven repository.

Note that a first attempt to do this was made in the now deleted branch jacoco, but due to some weird issue with a commit, Cirrus CI could no longer clone this branch. Cirrus CI seems to be using the library go-git, which has an obscure bug which sometimes prevent cloning from running (https://github.com/go-git/go-git/issues/495). Because of this issue, it was decided that it would be simpler to just recreate another branch and do a single clean commit, which seems to have fixed the issue with go-git.


Luca-Engel (approved):
Great job with this maven repository!

This will make it easier to get a sense of our branch coverage in the future and allow us to more quickly find out what parts of our code still need more testing.

PR #55, opened 2023-03-25
Mock db testing bugfix

Luca-Engel:
Since the signOut function of RealGoogleAuthenticator runs asynchronously and then calls an onComplete function, tests that do not take a certain timeout time into account will sometimes fail. This was the case for the dashboard test that checked if the sign-out button correctly redirects to the login screen (dashboardCorrectlyRedirectsOnLogOutClick).

Therefore, a check was added in the mock authenticator to then instantly execute the callback function. This is an easy fix for the problem that still keeps branch coverage the same as it was before.


app/src/androidTest/java/com/github/sdpcoachme/firebase/auth/MockAuthenticator.kt
```diff
@@ -19,7 +20,14 @@ class MockAuthenticator : Authenticator {
     }
 
     override fun signOut(context: Context?, onComplete: Runnable?) {
-        realAuthenticator.signOut(context, onComplete)
+        // Done since the tests for dashboard activity have no timeout
+        // capabilities to wait for the asynchronous sign-out to finish
+        // and, therefore, sometimes would fail
+        if (context is DashboardActivity) {
+            onComplete?.run()
```
brygotti, line 27: Doesn't this prevent the real authenticator to sign out the user when the activity is dashboard ? Since in this case the authenticator function isn't called.

brygotti (approved):
There probably is a cleaner way to solve this issue of "testing asynchronous functions where we can't access the onComplete callback", but it works well like this for now. We can still adapt it later on.

PR #57, opened 2023-03-27
Database and Selection of favorite sports

lindsaybordier:
The following will be added with this pull request:

- Refactor most of the database code, the Database interface, the FireDatabase Implementation an MockDatabase.
- Fix a an issue concerning return type of database methods, efficient refactoring to fix the issus in FireDatabase implementation with DataSnapshots.
- Code in other activities was adapted to take the new return type into account, as casting into maps was no needed anymore.
- Add method(s) to check wether a user is in the database.
- Add functionality for a user to select favorite sports after Signup and possibly from the profile editing activity. Favorite sports selection after signup redirects to Dashboard activity.
- Most of the work was done to allow further testing of the database implementation with the database emulator (see some commented code for that purpose). Some instructions in the cirrus yml file should allow for further testing in the CI if necessary in the future, also to improve coverage.


Luca-Engel (changes requested):
Great idea with the conversion of the database response to a UserInfo already in the FireDatabase class! This will make it much more convenient to use and also provides a great abstraction to the outside!

Before looking at the code, I tried running the application locally and spotted an error:
When the user tries to log in with an email that  is not yet linked to an account, instead of being redirected to the SignUpActivity, the error message "Unknown error" is displayed.

I then checked out the code in the LoginActivity and found that the following changes should fix the problem and at the same time refactor the existing code nicely:
I suggest refactoring the method launchPostLoginActivity to the following two  methods:
```
private fun launchPostLoginActivity(email : String) {
        signInInfo = email
        val query = database.getUser(email)
        query.handle { result, exception ->
            if (exception == null) {
                launchActivity(result.email, DashboardActivity::class.java)
            } else if (exception.cause is NoSuchKeyException) {
                launchActivity(email, SignupActivity::class.java)
            } else {
                signInInfo = "Unknown error"
            }
        }
    }

    private fun launchActivity(email: String, activity: Class<*>) {
        val intent = Intent(this, activity)
        intent.putExtra("email", email)
        startActivity(intent)
    }
```

The problem before was that exception.cause was not used and instead exception was used so the wanted branch was not taken. 

I also spotted a second error concerning the parsing of a UserInfo. I believe to have found the following fix:
```
data class UserInfo(
    val firstName: String = "",
    val lastName: String = "",
    val email: String = "",
    val phone: String = "",
    val location: String = "",
    val coach: Boolean = false,
    val sports: List<Sports> = emptyList()
) {
    // Constructor needed to make the data class serializable
    constructor() : this("", "", "", "", "", false, emptyList())
}
```
Since isCoach is serialized to coach when sending the information from and to the database (is... is assumed to be a boolean and, therefore, serialized without the "is" part) and we are now automatically converting the attributes, we need to adapt the name to "coach"




In the future, both running the tests and running the application manually will hopefully mitigate such errors.

lindsaybordier:
Thanks a lot for your constructive comments @Luca-Engel .
Changes made, everything seems to work fine now.

Luca-Engel (approved):
Thanks for the adaptations!

Now everything looks great and ready for merging.

PR #64, opened 2023-03-29
Week Schedule

Thymian18:
This PR will add the following features to the app:

- When a user clicks on the "Schedule" button of the navigation bar, the schedule activity is launched.
- Currently, during the onCreate function, some sampleEvents get added to the userInfo of the authenticated user in the database for illustrating/demo purposes.
- Those added events that are in the database get displayed in the schedule.
- At the moment, the schedule only shows the current week and it is not possible as a user to switch between weeks.
- Also, events can only be displayed at one day. So, multi-day events do not wrap around to the next day yet.
- The tests cover pretty much everything. One TODO is left for the next sprint that concerns error handling.

image


app/src/main/java/com/github/sdpcoachme/firebase/database/FireDatabase.kt
```diff
@@ -37,6 +38,17 @@ class FireDatabase(databaseReference: DatabaseReference) : Database {
...
     }
 
+    override fun addEventsToDatabase(email: String, events: List<Event>): CompletableFuture<Void> {
+        return this.getUser(email).thenAccept {
+            val updatedUserInfo = if (it.events.isEmpty()) {
+                it.copy(events = events)
+            } else {
+                it.copy(events = it.events + events)
+            }
+            addUser(updatedUserInfo)
```
lindsaybordier, line 48: The if else could probably be simplified into :
val updatedUserInfo = it.copy( it.events + events)

app/src/androidTest/java/com/github/sdpcoachme/ScheduleActivityTest.kt
```diff
@@ -0,0 +1,149 @@
...
+    fun addEventsToDatabaseUpdatesUserInfoCorrectly() {
+        defaultEmail
+        val oldUserInfo = database.getUser(defaultEmail)
+
+        database.addEventsToDatabase(defaultEmail, eventList)
+
+        val newUserInfo = database.getUser(defaultEmail)
+        newUserInfo.thenAccept {
+            assert(oldUserInfo != newUserInfo)
+            assert(it.events == eventList)
```
lindsaybordier, line 85: Maybe here using the assertTrue method of JUnit would be better.

In general, JUnit assertion methods provide more context when tests fail compared to kotlin buit-in assertions.

lindsaybordier (approved):
The code is great and I was able to run the app and test your Schedule activity locally, great job!!
The UI code is also nicely modularised which helped understanding this otherwise quite complex piece of code.

Just a few comments, most of which you already addressed (therefore I approve merging the code):

- Possible to simplify the code and remove the if else in the new database method to add events for a user
- Using JUnit asserts instead of kotlin built-in asserts to provide more context when tests fail
- (not yet addressed) The current hardcoded example events are duplicated in the Schedule activity and its test class, which I believe is not necessary. However, these are temporary to allow for showing functionality during the demo, so I guess we can remove them during the next sprint when more functionality is implemented to use the schedule.

PR #70, opened 2023-03-30
Display coach profile to client

Luca-Engel:
This PR adds a feature that allows the user to view the profile of a coach.

Now, when the user clicks on one of the listed coaches in the CoachesListActivity, the user is redirected to ProfileActivity (used to be EditProfileActivity) where that coach's profile is shown. Alongside the profile, there is a "message button" that, in the future, can be used to message that coach.

Tests for this Feature have also been added and adapted.

Additionally, in CoachesListActivityTest, two tests have been adapted to cover more test cases.


Thymian18 (approved):
LGTM, very clean changes and cool new feature to see the coach's profile. Regarding the tests, they seem to cover the big part of the code and they make sense. Also, the sports icons add some additional spice to the app. Very good work, keep it up:)

PR #84, opened 2023-04-04
Update google maps key used in Cirrus

brygotti:
Change the key so that it is the same as the one we personally use in our local repos.

This allows us to have a single key for all the Google Maps queries, coming either from our local repo or from the Cirrus CI, and add billing information linked specifically to that key.


Ayskin (approved):
Nice, so now we have everything set up and done concerning our key management! This will enable us to manage the whole project through one account, so we are now free to set any restrictions we'd like.

PR #86, opened 2023-04-09
Client side cache

lindsaybordier:
This PR consists of the first 2 of my assigned sprint tasks:
- Implement client-side caching for database operations
- Handle global email differently so it does not have to be passed through an extra when switching activities.

1) Client Side Caching:
Should be pretty straight forward. Caching Database constructor takes a wrapped database instance as parameter.
Tried to implement it as modular as possible. Used with the Mockdatabase in the tests.

2) Email refactoring:
This is a global refactoring, most of the code base was adapted.
The current user email is now a private attribute of the database instance with a setter and a getter.
Important pattern change: the check for null email that was done before in each activity now checks if email == "" to redirect  to the error page, as the current user email is never null.
All tests were adapted and some checks for the email extra were removed.
Tried to remove code duplication in tests every time there was an opportunity.


Ayskin (changes requested):
Clear and nice implementation of client-side caching.

In addition to the code comments, I suggest the following changes before merging into main :
- we can make this caching effective by wrapping our application firebase with it
- for the new pattern `email == ""`, we can use `email.isEmpty()` instead for better readability and for some sort of assertion because the email should now never be null
- handle the conflict explained below

With this new email management, when we click on a nearby coach profile, we get the current user profile. This happens because the `ProfileActivity` does not receive an email anymore when launched from an Intent. 

As the target email is still sent along with the intent in `CoachesListActivity` (see attached image), we can rollback the change made to `ProfileActivity` and allow the activity to receive an extra string as before.
Screenshot from 2023-04-11 13-09-36

Also be aware that the test `whenClickingOnACoachProfileActivityShowsCoachToClient()` in `CoachesListActivityTest` for now sets the currentEmail to the target coach which breaks the wanted behavior. Some tests may therefore need to be slightly changed accordingly.

lindsaybordier:
Thanks a lot for the review and your comments @Ayskin 
The concerns were all addressed. Please note the following:

- Caching will now be used in production
- isEmpty is now used for checks
- I removed the get and set methods from the Database interface and its implementations. These methods were not relevant to the application and could cause problems if used badly in code. For example: giving a certain path to update a user using the "set" method instead of using the relevant "updateUser" method could leave the cache in an inconsistent state and cause problems. Therefore it was better to remove those methods from the interface.
- The ProfileActivity is now fixed as well as its tests. It now checks if the user is viewing a coach or its own profile, and based on that it either uses the user email (database instance) or the target coach's email (sent as an extra along with isViewingCoach).

PR #87, opened 2023-04-13
Schedule eventwrapping

Thymian18:
In this PR, I implemented the following schedule extension:
Whenever an event is not only affecting one day but multiple days, the event is now correctly displayed on all days. Before, the event was only displayed on the first day until midnight but not on all following days.
To get to the goal, I added a type ShownEvent so that I have two types for events now. The type Event is the one that is saved in the DB, the type ShownEvent is the one that is used to actually display the event. Additionally to the parameters of Event, ShownEvent also takes startText and endText so that we can differentiate between the times an event spans over and the times displayed as text.
In case an event spans multiple days, it corresponds to one Event instance which is converted to multiple ShownEvent instances (one per day). 

Let's say, we have an event that starts on 13.04.2023 at 18:00 and ends on 15.04.2023 at 06:00. Then, one instance with those start and end parameters is stored in the DB. Since the event affects 3 days, there are going to be 3 instances of type ShownEvent, each having start and end adjusted but the same startText "2023-04-13T18:00" and endText "2023-04-15T06:00".


app/src/androidTest/java/com/github/sdpcoachme/schedule/EventOpsTest.kt
```diff
@@ -0,0 +1,94 @@
...
+    )
+
+    @Test
+    fun eventsToWrappedEventsCreatesCorrectMap() {
+        EventOps.eventsToWrappedEvents(eventList)
+        val actualMap = EventOps.getMultiDayEventMap()
+        val expectedMap = mutableMapOf<Event, List<ShownEvent>>()
+
+        eventList.forEach { event ->
+            println("Event: ${event.name}")
```
Luca-Engel, line 69: I would say, the println's can be removed

app/src/androidTest/java/com/github/sdpcoachme/schedule/ScheduleActivityTest.kt
```diff
@@ -63,13 +69,13 @@ class ScheduleActivityTest {
             end = currentMonday.plusDays(3).atTime(15, 0, 0).toString(),
             description = "Learn about the latest design improvements to help you build personal dynamic experiences with Material Design.",
         ),
-        Event(
+        /*Event(
```
Luca-Engel, line 72: Is this commented out intentionally? If yes, I would recommend to delete it to increase readability

app/src/main/AndroidManifest.xml
```diff
@@ -54,7 +54,7 @@
             android:label="@string/title_activity_signup"
             android:theme="@style/Theme.CoachMe" />
         <activity
-            android:name=".ScheduleActivity"
+            android:name=".schedule.ScheduleActivity"
```
Luca-Engel, line 57: great idea of moving this activity into a separate package, this makes it much easier to get an overview of the project parts!

brygotti (changes requested):
Putting activities in packages as you did makes things much cleaner. Please consider changing the testing code as proposed in the comment. Other than that, it's all good for me!

Luca-Engel (changes requested):
Great idea of storing the actual events as is in the db and converting them to shown events locally. This way, you can easily get an overview of the Events in the db and can use those events for future feature implementations (such as creating events as a user).

There are some minor code deletions that can be done as mentioned in the comments, other than that, great job!

app/src/main/java/com/github/sdpcoachme/schedule/EventOps.kt
```diff
@@ -0,0 +1,112 @@
...
+import java.time.temporal.ChronoUnit
+
+class EventOps {
+    companion object {
+        /**
+         * A map to keep track of events that span multiple days. Has to be changed once the events are modified.
+         */
+        private val multiDayEventMap = mutableMapOf<Event, List<ShownEvent>>()
+
+        fun getMultiDayEventMap() = multiDayEventMap
```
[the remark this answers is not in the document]
Thymian18: For now, the function is indeed only used in the testing. It is going to be necessary however to have that function for the future when I add the ability to let a user add and modify events.

brygotti (approved):
LGTM

Luca-Engel (approved):

PR #88, opened 2023-04-14
Location autocomplete in signup, and refactoring

brygotti:
This PR introduces the usage of the Google Places API to store and select locations in the app. That way, we make sure the user can only select a location that is recognized by Google Maps.

1. During signup, the user now needs to select a location using Google Places Autocomplete activity.
2.  Instead of storing location as String in the database, we store UserLocation objects, that describe a Place from Google Places API. In the database, we cache the id, the full address and the GPS coordinates of the Place, so that we don't need to request it to Google Places API everytime we need those info.
3. Since the Google Places Autocomplete activity is hard to test (tests require UI Automator and a lot of tricks, which seems to make them undeterministic), we mock the launch of the activity and bypass it in the tests (see MockLocationAutocompleteHandler)
4. In CoachesListActivity, display coaches ordered by how close they are from current location. For now, they are sorted based on a hard-coded fixed location, but this is because some refactoring will be done to the app's map interface, and we wait for those to be completed before using it. A TODO was left for that purpose.
5. SignupActivity test was refactored and is now more modular, and uses the mock handler for the location autocomplete
6. All other tests were refactored to use UserLocation instead of strings in the UserInfo that were created. A class LocationSamples was added to easily use UserLocation without needing to recreate them in each test.
7. Regarding the database, some methods were modularized similarly to what was done in main, and a method was added to retrieve all users of the app at once. This is clearly not efficient, but is needed to sort the users based on how close they are from current location. Indeed, we can't run complex queries in Firebase, so the only option is to do this. The Database interface was also augmented with new methods to get all users and get them ordered by distance.
8. Some small bugs and TODOs were fixed in CoachesListActivity, SignupActivity and FireDatabase.

Note: for now, the ProfileActivity in edit mode does not allow to update the user location. This is because it will be refactored in the next sprint, and it was decided that it was not useful to add this functionality now and refactor it right after.


app/src/main/java/com/github/sdpcoachme/CoachMeApplication.kt
```diff
@@ -14,12 +19,19 @@ import com.google.firebase.ktx.Firebase
 open class CoachMeApplication : Application() {
     // For DI in testing, add reference to dependencies here
     open lateinit var database: Database
-    open lateinit var email: String
```
lindsaybordier, line 17: This was indeed not necessary anymore, great catch.

app/src/main/java/com/github/sdpcoachme/firebase/database/Database.kt
```diff
@@ -29,6 +31,26 @@ interface Database {
...
+                val userLatitude = user.location.latitude
+                val userLongitude = user.location.longitude
+                val distance = SphericalUtil.computeDistanceBetween(
+                    LatLng(latitude, longitude),
+                    LatLng(userLatitude, userLongitude)
+                )
+                distance
+            }
+        }
+    }
```
lindsaybordier, line 52: Great code, very readable.

app/src/main/java/com/github/sdpcoachme/firebase/database/FireDatabase.kt
```diff
@@ -52,14 +51,31 @@ class FireDatabase(databaseReference: DatabaseReference) : Database {
...
      * @param databaseChild the database reference in which to set the key-value pair
      * @param key the key of the value
      * @param value the value to set
      * @return a completable future that completes when the child is set
+     * @throws IllegalArgumentException if the key is empty (no child can have an empty key)
      */
     private fun setChild(databaseChild: DatabaseReference, key: String, value: Any): CompletableFuture<Void> {
+        // Careful here, giving an empty string as a key will return a reference to the parent!!
+        if (key.isEmpty())
+            throw IllegalArgumentException("Key cannot be empty")
```
lindsaybordier, line 78: This is a very important modification, as giving an empty key was triggering a much wrong behaviour before, instead of throwing an exception.
Nice.

app/src/main/java/com/github/sdpcoachme/location/autocomplete/LocationAutocompleteHandler.kt
```diff
@@ -0,0 +1,33 @@
...
+     */
+    fun launch(): CompletableFuture<UserLocation>
+
+    // Used to handle places autocomplete activity errors
+    class AutocompleteFailedException(message: String? = null, cause: Throwable? = null) : Exception(message, cause) {
+        constructor(cause: Throwable) : this(null, cause)
+    }
+    class AutocompleteCancelledException(message: String? = null, cause: Throwable? = null) : Exception(message, cause) {
+        constructor(cause: Throwable) : this(null, cause)
+    }
```
lindsaybordier, line 32: The constructors for those classes don't seem necessary and are never used, maybe not worth keeping them ?

app/src/main/java/com/github/sdpcoachme/CoachesListActivity.kt
```diff
@@ -28,23 +28,30 @@ import androidx.compose.ui.text.style.TextOverflow
...
 
-            // TODO: Need to handle the future correctly in cases like this (might need to use coroutines)
-            var f by remember { mutableStateOf(futureListOfCoaches) }
-            f.thenAccept {
-                listOfCoaches = it
-                f = CompletableFuture.completedFuture(null)
+            LaunchedEffect(true) {
+                listOfCoaches = futureListOfCoaches.await()
+                stateLoading.complete(null)
             }
```
lindsaybordier, line 55: I would put a small comment before LaunchedEffect to explain the purpose for testing.

lindsaybordier (changes requested):
This is great code and much refactoring of the code base was done which definitely helps keeping it clean.
The modularisation is top notch: even though some mechanisms are quite complex they are made clearly readable.
Also, the solution of having public futures in activities to wait for specific task to finish is a great way to remove some of the races we had in the tests even though they were passing.

  Please take a look at some minor comments:
- Removing constructors in LocationAutoCompleteHandler
- Add comment for code used in testing framework in CoachesListActivity

When running the application locally, I noticed that trying to display nearby coaches makes the application crash.
This is because the old location entries in the database were strings and can not be de-serialised as UserLocation objects. Therefore cleaning the database is necessary to be able to use this functionality again.

app/src/main/java/com/github/sdpcoachme/location/autocomplete/LocationAutocompleteHandler.kt
brygotti: Those constructor are here to match exactly the interface of the Exception class (which is a Java class). See https://stackoverflow.com/questions/45162869/kotlin-throw-custom-exception for more info about why it is a good idea to keep them.

brygotti:
We will indeed need to update the database. I created a new issue for this: #90.

brygotti:
Note: I added a new commit, but the commit message is wrong. I don't want to mess with the git history, so I just added a comment in the commit to precise what the correct message was.

lindsaybordier:
Ok, LGTM. Thanks!

lindsaybordier (approved):

PR #91, opened 2023-04-15
In app messaging

Luca-Engel:
The following pull request will add the in-app messaging feature. This feature contains the following:

- CoachesListActivity has been refactored to not only be able to show the list of nearby coaches when clicking on that button in the dashboard but also allow the displaying of the contacts to the client when clicking on the messaging button in the DashboardActivity
- messaging button click in the coach viewing section (nearby coaches) takes the user to a chat with that coach so that the user can contact him through the messaging feature
- a messaging section has been added to the dashboard so that the user can see a view of all their contacts that they have messaged with. From that list, the user can then go to the chat with a given other user by clicking on their name
- the messaging interface is straight forward and designed with other messaging apps' interfaces as a template to allow the user to see a familiar design (see the screen shot at the bottom)
- the messaging interface contains a clickable header that takes the current user to the profile of the other user, a back button to go back to the contacts, a scroll-to-bottom button is shown whenever the chat is not scrolled all the way down to show to the user that there are more messages further down.
- the chat interface updates the displayed messages whenever a new message appears in the db. This is done with a ValueEventListener.
- similar to WhatsApp and other messaging apps, every message contains a timestamp and the messages sent by the current user are annotated with a check mark when they are sent and with a blue double check mark when they have been read by the other user.
- Lastly, inside the chat, the date is displayed of when the messages have been sent as in other messaging apps (see the screen shot)

SDP Image of Chat
SDP Image of Chat 2


app/src/androidTest/java/com/github/sdpcoachme/CoachesListActivityTest.kt
```diff
@@ -40,7 +40,9 @@ class CoachesListActivityTest {
 
     // With this, tests will wait until activity has finished loading state
     @Before
-    fun setup() {
+    open fun setup() {
+        println("outer setup")
```
Thymian18, line 44: Might be better to delete that print statement.

app/src/androidTest/java/com/github/sdpcoachme/CoachesListActivityTest.kt
```diff
@@ -111,6 +113,51 @@ class CoachesListActivityTest {
             Intents.release()
     }
 
+    // Subclass added to be able to run a different setup method (to simulate viewing contacts)
+    class ContactsListTest: CoachesListActivityTest() {
+        @Before
+        override fun setup() {
+            // Launch the activity
+            println("Launching activity in inner setup")
```
Thymian18, line 121: Eventually delete that print too

app/src/androidTest/java/com/github/sdpcoachme/CoachesListActivityTest.kt
Thymian18, line 1: As mentioned, delete the prints if they are not absolutely necessary (which I think is the case). Otherwise, all good.

app/src/androidTest/java/com/github/sdpcoachme/messaging/ChatActivityTest.kt
```diff
@@ -0,0 +1,315 @@
...
+        }
+    }
+
+    @Test
+    fun whenScrolledToTheBottomScrollButtonIsNotDisplayed() {
+        val chatIntent = Intent(ApplicationProvider.getApplicationContext(), ChatActivity::class.java)
+        chatIntent.putExtra("toUserEmail", toUser.email)
+
+        ActivityScenario.launch<ChatActivity>(chatIntent).use {
+            composeTestRule.onNodeWithTag(SCROLL_TO_BOTTOM, useUnmergedTree = true).assertDoesNotExist()
```
Thymian18, line 105: Maybe, there exists a possibility to perform a scroll down first before doing this check, just to be sure that it is actually at the bottom. I assume it always starts at the bottom, but this way the test would be a tiny bit more convincing.

app/src/androidTest/java/com/github/sdpcoachme/messaging/ChatActivityTest.kt
```diff
@@ -0,0 +1,315 @@
...
+        ActivityScenario.launch<ChatActivity>(chatIntent).use {
+            database.addChatListener("run-previous-on-change") {}
+
+            composeTestRule.onNodeWithText("test onChange method", substring = true, useUnmergedTree = true)
+                .assertIsDisplayed()
+        }
+    }
+
+    @Test
+    fun messageSentByOtherUserDoesNotContainIsReadCheckMark() {
```
Thymian18, line 232: Very good that you thought of this test case.

app/src/androidTest/java/com/github/sdpcoachme/messaging/ChatActivityTest.kt
Thymian18, line 1: Very extensive test file with all necessary tests. The naming of the tests makes clear what the goal of the test is which makes the tests very well-comprehensible.

app/src/main/java/com/github/sdpcoachme/data/UserInfo.kt
```diff
@@ -11,8 +11,9 @@ data class UserInfo(
     val location: UserLocation,
     val coach: Boolean,
     val sports: List<Sports> = emptyList(),
-    val events: List<Event> = emptyList()
+    val events: List<Event> = emptyList(),
+    val chatContacts: List<String> = listOf(), // TODO: order based on message timestamp
```
Thymian18, line 15: Check if this TODO comment can be removed.

app/src/main/java/com/github/sdpcoachme/firebase/database/FireDatabase.kt
```diff
@@ -51,6 +64,83 @@ class FireDatabase(databaseReference: DatabaseReference) : Database {
         currEmail = email
     }
 
+    override fun getChatContacts(email: String): CompletableFuture<List<UserInfo>> {
+        return getUser(currEmail).thenApply {
+            it.chatContacts
+        }.thenCompose { list ->
+            val mappedF = list.map { email ->
+                getUser(email) //TODO: how to change this to not use .get() ???
```
Thymian18, line 72: Again, the TODO comment can be removed I guess.

app/src/main/java/com/github/sdpcoachme/data/UserInfo.kt
Luca-Engel: Thank you for spotting this. Indeed, this todo is a little unclear but this feature has been implemented (it refers to ordering the contacts so the the most recently messaged contacts come first). I will delete it.

app/src/androidTest/java/com/github/sdpcoachme/messaging/ChatActivityTest.kt
Luca-Engel: Thank you for this comment. As the chat's scroll starts all the way at the bottom of the chat, the scroll-button should not be displayed when launching this activity. I will add a comment to the test to clarify it.

app/src/main/java/com/github/sdpcoachme/messaging/ChatActivity.kt
Thymian18, line 1: Such a big amount of code and lots of cool features that elevate the user experience, but still very clear abstraction with all the sub-composables.

Thymian18 (changes requested):
This messaging feature is a really big extension to our app and you made great progress. There are only some comments that can/should be removed. Other than that, awesome work and extensive testing.

Luca-Engel:
Thank you for your detailed code review! The requested changes have now been addressed.

Luca-Engel:
Thank you for your detailed review! The requested changes have now been addressed.

Thymian18 (approved):
LGTM

PR #92, opened 2023-04-16
Message coach button click from profile test

Luca-Engel:
This small pull request completes a started test from the last sprint now that the in-app messaging has been implemented and refactors todos
- The test checks if the "message coach" button redirects the user to the ChatActivity. The corresponding todos have been removed.
- Also, this pr specifies that the added todos for the caching database should be implemented in the next sprint to enable caching for the messaging part of CoachMe.


Thymian18 (approved):
LGTM

PR #93, opened 2023-04-17
Dashboard refactored

Ayskin:
This branch corresponds to the task of refactoring the dashboard, in order to grant access to it from all other activities of the application. Activities layouts were adapted to integrate the application bar of the dashboard.

The branch also consists in some code refactoring to reduce code duplication and fixes some wrong test names or test behaviors.

The DashboardActivity becomes the Dashboard UI that is called in each activity with a specific content related to the current launched activity.

MapActivity is the new starting activity that holds the main map. It is also set as a new entry in the dashboard to be accessible as other activities.

The MapViewModel was removed. The location retrieval pipeline was entirely moved into the MapActivity. In addition, the userLocation is now linked to the application as for the database. So keep in mind that it is now accessible from everywhere in the application.
This enabled for example to get the user location for the nearby coaches sorting, which fulfills one core functionality of the app !

This branch also includes test refactoring to cope with the new dashboard implementation and also new MapActivity tests.


app/src/main/java/com/github/sdpcoachme/CoachesListActivity.kt
```diff
@@ -43,77 +42,66 @@ class CoachesListActivity : ComponentActivity() {
...
 
-            // Proper way to handle result of a future in a Composable.
-            // This makes sure the listOfCoaches state is updated only ONCE, when the future is complete
-            // This is because the code in LaunchedEffect(true) will only be executed once, when the
-            // Composable is first created (given that the parameter key1 never changes). The code won't
-            // be executed on every recomposition.
-            // See https://developer.android.com/jetpack/compose/side-effects#rememberupdatedstate
-            LaunchedEffect(true) {
-                listOfCoaches = futureListOfCoaches.await()
+        if (email.isEmpty()) {
```
lindsaybordier, line 63: I think it would make more sense to check if email is empty before using it at line 50 in futureListOfCoaches. As we check it, better do it as early as possible. (even though an empty email given to getChatContacts will trigger an IllegalArgumentException)

app/src/androidTest/java/com/github/sdpcoachme/schedule/ScheduleActivityTest.kt
```diff
@@ -201,4 +163,20 @@ class ScheduleActivityTest {
             }
         }
     }
+
+    @Test
+    fun dashboardHasRightTitleOnSchedule() {
+        ActivityScenario.launch<ScheduleActivity>(defaultIntent).use {
+            composeTestRule.onNodeWithTag(BAR_TITLE).assertExists().assertIsDisplayed()
+            composeTestRule.onNodeWithTag(BAR_TITLE).assert(hasText("Schedule"))
```
lindsaybordier, line 171: It would be better to use assertTrue from JUnit as more info is printed in case the check fails.

app/src/androidTest/java/com/github/sdpcoachme/messaging/ChatActivityTest.kt
```diff
@@ -85,8 +91,7 @@ class ChatActivityTest {
 
     @Test
     fun startingElementsArePresent() {
-        val chatIntent = Intent(ApplicationProvider.getApplicationContext(), ChatActivity::class.java)
-        chatIntent.putExtra("toUserEmail", toUser.email)
+        val chatIntent = defaultIntent.putExtra("toUserEmail", toUser.email)
```
lindsaybordier, line 94: This line is duplicated in all tests except errorHandlerIsLaunchedIfToUserEmailNotPassedInIntentExtra so I would recommend to put the extra (email) directly when declaring the defaultIntent at line 51 and create a special intent without email for the specific test just mentioned.

app/src/androidTest/java/com/github/sdpcoachme/map/MapActivityTest.kt
```diff
@@ -0,0 +1,144 @@
...
+            composeTestRule.onNodeWithTag(GO_TO_LOGIN_BUTTON).assertIsDisplayed()
+            composeTestRule.onNodeWithTag(ERROR_MESSAGE_FIELD).assertIsDisplayed()
+        }
+    }
+
+    @Test
+    fun dashboardHasRightTitleOnMap() {
+        ActivityScenario.launch<MapActivity>(defaultIntent).use {
+            composeTestRule.onNodeWithTag(BAR_TITLE).assertExists().assertIsDisplayed()
+            composeTestRule.onNodeWithTag(BAR_TITLE).assert(hasText("Coach Me"))
```
lindsaybordier, line 122: Better to use assertTrue

lindsaybordier (changes requested):
Great code!

Please take a look at a few minor comments directly on the code.

As a more general comment:
The userLocation is declared in CoachMeApplication and used as a variable accessible from all classes which is completely valid and great modularisation. However, when refactoring the email (that was passed as an extra to activities before) we discussed with the team that we would store currentUser variables directly in the database object (as an attribute with a setter and a getter). This is to avoid declaring many variables in CoachMeApplication and make them easily accessible from the database instance that is already accessible everywhere.
Therefore it would make more sense to have the userLocation as an attribute of the database object (eg currentUserLocation) just like for the current user email.

If you do not have time to change this we can always do it in a next sprint or re-discuss the architecture in the next meeting.

brygotti:
Just created #94 to handle this issue

app/src/androidTest/java/com/github/sdpcoachme/schedule/ScheduleActivityTest.kt
Ayskin: The assert function comes along with the compose UI tests library and is usefull here to call the hasText(..) matcher. As we work with the compose SemanticsNode here, it is more natural to use these functions.

app/src/androidTest/java/com/github/sdpcoachme/messaging/ChatActivityTest.kt
Ayskin: I've modified this, thanks for the comment.

lindsaybordier (approved):

PR #96, opened 2023-04-19
Keyboard next

lindsaybordier:
This pull request corresponds to adding the keyboard functionality of a next button that automatically takes the user to the next field to fill out forms more smoothly.

This functionality was intended for signup, but I extended it to the editing of the profile activity.

Each time the user enters text in a field, a next button is made available on the keyboard. Pressing this button changes the focus to the next text field or clears out the focus (close the keyboard) if the user was on the last field.

Text filling was more robust, allowing to only enter one line for each field.

After making tests for this functionality, new tests were added for the cachingDatabase to improve the test coverage of the project.


app/src/androidTest/java/com/github/sdpcoachme/CachingDatabaseTest.kt
```diff
@@ -33,44 +38,144 @@ class CachingDatabaseTest {
...
+                cachingDatabase.updateUser(defaultUser) }
             .thenCompose { cachingDatabase.getUser(exampleEmail) }
             .get(5, SECONDS)
         assertTrue(cachingDatabase.isCached(exampleEmail))
-        assertEquals(newUser, updatedUser)
+        assertEquals(defaultUser, updatedUser)
+    }
+
+    @Test
+    fun addAllUsersPutsAllUsersInCache() {
```
Thymian18, line 63: I am not sure if I understood what you want to test here. I think it makes sense to test those things, but rather in two separated tests. 
We can have one test where you create the wrappedDatabase, then add all the users to it, then create the cachingDatabase, and then call getAllUsers and check if the users are cached. This test will test the caching functionality of getAllUsers.
As a second test, you can check if calling getAllUsers caches the users correctly.

app/src/androidTest/java/com/github/sdpcoachme/CachingDatabaseTest.kt
```diff
@@ -33,44 +38,144 @@ class CachingDatabaseTest {
...
+    fun setAndGetCurrentEmail() {
+        val wrappedDatabase = MockDatabase()
+        val cachingDatabase = CachingDatabase(wrappedDatabase)
+        val email = "test@email.com"
+        cachingDatabase.setCurrentEmail(email)
+        assertEquals(email, cachingDatabase.getCurrentEmail())
+    }
+
+    @Test
+    fun addEventsToUserPutsUserInCacheAndUpdatesEvents() {
```
Thymian18, line 86: Here again, maybe this test can be split up into two tests.

Thymian18 (changes requested):
Overall, good work, just consider the two comments I left for the tests of the cachingDatabase.

app/src/androidTest/java/com/github/sdpcoachme/CachingDatabaseTest.kt
lindsaybordier: Actually here the goal is to test if the getAllUsers method puts all the users in the cache. For that, we first have to put all users in the database and clear the cache.
Another test checks if the method updateUser puts the user in the cache.

Therefore, I changed this test's name to getAllUsersPutsAllUsersInCache

lindsaybordier:
Thanks for the review! Comments were addressed.

Thymian18 (approved):
LGTM

PR #98, opened 2023-04-20
Add markers on the map to see nearby coaches

brygotti:
This PR adds markers to the map that indicate where nearby coaches are located. For now, given the (very) small userbase of our app, we display all coaches on the map, but for better perfomance, we might decide later to only display coaches close to the camera position of the map, or close to the position of the user.

When clicking on a marker, a pop-up window opens showing quick information about the coach. One can then click on this window to open the full-page profile of the coach in a new activity.

After trying multiple times to test the existence of markers on the map without success (it seems they are not registered in compose' semantics node tree, but I couldn't figure out how to make them detectable in the tree), I decided to keep the testing code (tags, completablefuture for UI loading, ...) but to comment out the tests. That way, if we ever decide to test the markers, we can further investigate the issue without having to start from scratch.

### DIFF
--- a/app/src/androidTest/java/com/github/sdpcoachme/SignupActivityTest.kt
+++ b/app/src/androidTest/java/com/github/sdpcoachme/SignupActivityTest.kt
@@ -26,6 +26,7 @@ import org.junit.Before
 import org.junit.Rule
 import org.junit.Test
 import org.junit.runner.RunWith
+import java.util.concurrent.CompletableFuture
 import java.util.concurrent.TimeUnit.SECONDS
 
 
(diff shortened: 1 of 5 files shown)


app/src/androidTest/java/com/github/sdpcoachme/SignupActivityTest.kt
```diff
@@ -106,16 +107,18 @@ open class SignupActivityTest {
...
+        lateinit var databaseStateSending: CompletableFuture<Void>
+        scenario.onActivity {
+            databaseStateSending = it.databaseStateSending
         }
+        var exceptionThrown = false
+        databaseStateSending.handle { _, _ ->
+            exceptionThrown = true
+            // Recover from exception
+            null
+        }.get(10, SECONDS)
```
Thymian18, line 119: Yes, for all those get calls, it is important to have them outside to make sure there are no deadlocks.

app/src/main/java/com/github/sdpcoachme/map/MapActivity.kt
```diff
@@ -170,5 +212,74 @@ fun Map(modifier: Modifier, lastUserLocation: MutableState<LatLng?>) {
...
+                mutableStateOf(
+                    MarkerState(LatLng(user.location.latitude, user.location.longitude))
+                )
+            }
+
+            MarkerInfoWindowContent(
+                state = state,
+                tag = MARKER(user),
+                onInfoWindowClick = {
+                    // TODO: code similar to CoachesList, might be able to modularize
```
Thymian18, line 227: We should definitely take a look at that code part in the next sprint to refactor common code parts. But for now, I think it is fine.

Thymian18 (approved):
As I did the map bootcamp, I know how hard the testing for that part is. It really seems to be a problem. Hard to tell the functionality of the markers' implementation but since tested locally, it should be good.
One thing that will possibly be doable next sprint is to have the common code part in map activity refactored.
The work with all the futures does not seem that easy, good job.

Issue #89, opened 2023-04-14
When editing the profile, selecting sports doesn't update the profile activity


brygotti:
When editing the profile, selecting sports doesn't update the profile activity directly, we need to refresh the activity once before the changes appear.

PR #112, opened 2023-04-23
Refresh of profile activity

### LINKED ISSUE #95: Seeing a coaches profile still writes "My profile" in the dashboard title

brygotti:
This PR contains a nearly complete rewrite of ProfileActivity to make it more user-friendly. Information is presented in rows, and clicking on a row opens an auxiliary activity that allows to edit the given field. For the location field, the location autocomplete activity opens up, for the sports field, it is the SelectSportsActivity, and for text valued fields, it is a newly created EditTextActivity that shows a text field and allows to edit it. To see how EditTextActivity works, please take a look at the javadoc.

Moreover, when the activity is launched with isViewingCoach set to true, the rows are no longer editable, but instead clicking on a row opens an activity to handle the given field: clicking on the coach's email launches the email app and creates a new draft with the email address set as the recipient address, clicking on the coach's location opens the location in the Google Maps app, and clicking on the phone number opens the phone app and dials the number.

Lastly, all activities have been moved to packages to have a clearer structure of the app, as adviced by the SDP coaches in their last feedback.

All ProfileActivity tests were updated, some new tests were added for new behaviors like launching external apps, and tests for EditTextActivity were added too. Given the restructuring of packages, some tests no longer passed in the CI, which seems to be related to some nondeterministic behavior of the tests depending on the order of test runs. Those tests were fixed. Note that the fixes are not ideal but they would require a big refactor of the testing codebase to be made properly, which is not a priority as of now.

This fixes #95 but not #89 given that #89 requires refactoring SelectSportsActivity.


app/src/androidTest/java/com/github/sdpcoachme/data/UserInfoSamples.kt
```diff
@@ -0,0 +1,54 @@
+package com.github.sdpcoachme.data
+
+class UserInfoSamples {
```
Luca-Engel, line 3: Nice refactoring idea! Now, we can mitigate unnecessary code duplications for creating users

app/src/androidTest/java/com/github/sdpcoachme/profile/EditTextActivityTest.kt
```diff
@@ -0,0 +1,41 @@
...
+        ActivityScenario.launch<EditTextActivity>(intent).use {
+            composeTestRule.onNodeWithTag(MAIN).assert(hasText(initialValue))
+            composeTestRule.onNodeWithTag(TITLE).assertTextEquals(title)
+            composeTestRule.onNodeWithTag(MAIN).performTextClearance()
+            composeTestRule.onNodeWithTag(MAIN).performClick()
+            composeTestRule.onNodeWithText(placeholder).assertIsDisplayed()
+            composeTestRule.onNodeWithText(label).assertIsDisplayed()
+        }
+    }
+
```
Luca-Engel, line 39: Nice job with the edit text activity. This will make the user experience a lot nicer!

Just a little idea for an additional test:
I believe that the default values are not tested for this activity (e.g., where initialValue is set to "", title is set to DEFAULT_TITLE, etc.). By adding a test for this case, we could increase test coverage a little.

Luca-Engel (changes requested):
Overall, I find it a great way how you implemented this refactoring! I believe the user experience becomes much nicer now that the profile editing has a more unified program flow.

I agree with your comment about it being hard to test the result of EditTextActivity. I also believe that it is smarter to test that only later if we need to increase coverage instead of spending an unnecessary large amount of time on it now.

I have still added a comment suggesting to add a test in EditTextActivityTest to increase test coverage by also testing if the default values are displayed correctly. 

Once this case is covered, this pull request will be ready to merge, great job!

app/src/androidTest/java/com/github/sdpcoachme/profile/EditTextActivityTest.kt
```diff
@@ -0,0 +1,41 @@
...
+    // Very hard to test output returned by the activity in tests. For now, we don't test it.
```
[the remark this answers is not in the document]
brygotti: Yes. It must be noted that activities that work based on startActivityForResult (like EditTextActivity) cannot be well tested without using a "parent" activity that launches them and handles their result. Given that it seemed out of proportion to create a whole new activity (which would need to live in "main" source folder and not in "androidTest") for the sole purpose of testing EditTextActivity, I decided not to test it for now. We do still have basic testing of its functionnalities in ProfileActivity which launches editTextActivity.

brygotti:
Thanks. Two more tests have been added as suggested.

Issue #65, opened 2023-03-29
Bug, network operation is launched on main ui thread


Ayskin:
The tests userInfoSelectedSportCorrectlyReplaced and userInfoUpdatedWithAllSelectedSportsAndRedirectedToDashboardActivity of SelectSportActivityTest crash with a Google Map api exception : Not on main thread. Bug showed up after merging the to the main added profile editing sport selection into the main-map branch that implements the map view.

Ayskin:
see #80 now for more information !!

PR #113, opened 2023-04-23
Fix for select sports bug

### LINKED ISSUE #89: When editing the profile, selecting sports doesn't update the profile activity

brygotti:
This PR fixes #89, and refactors the SelectSportsActivity to be compatible with the same interface as EditTextActivity.

Tests were also updated accordingly.


brygotti:
Note that #65 was not fixed here, but is no longer an issue in SelectSportsActivityTest, it is now in SignupActivityTest that it appears.

app/src/androidTest/java/com/github/sdpcoachme/profile/ProfileActivityTest.kt
```diff
@@ -303,6 +293,49 @@ class ProfileActivityTest {
...
+            composeTestRule.onNodeWithTag(SPORTS, useUnmergedTree = true).performClick()
+
+            for (rowTag in ROW_TEXT_LIST) {
+                composeTestRule.onNodeWithTag(rowTag.ROW, useUnmergedTree = true).performClick()
+            }
+            composeTestRule.onNodeWithTag(SelectSportsActivity.TestTags.Buttons.DONE, useUnmergedTree = true).performClick()
+
+            waitForUpdate(it)
+            // Given that we click on all sports, the list of sports is the complement
+            for (sport in Sports.values().toSet() - NON_COACH_2.sports.toSet()) {
```
Thymian18, line 311: As I understand it, Sports.values() contains all possible sports, but NON_COACH_2.sports at this point also contains all sports since ticked everything at line 305. Therefore, the for loop will do nothing. Am I wrong at this?

app/src/androidTest/java/com/github/sdpcoachme/profile/ProfileActivityTest.kt
```diff
@@ -303,6 +293,49 @@ class ProfileActivityTest {
         }
     }
 
+    @Test
+    fun editSports() {
```
Thymian18, line 297: Maybe, you could change the name of the test to something a bit more meaningful e.g., "choosingAllSports..."

app/src/androidTest/java/com/github/sdpcoachme/profile/ProfileActivityTest.kt
brygotti: UserInfo is an immutable data class, meaning it will be replaced in the database with a new instance with updated sports. That means that NON_COACH_2 still contains the old sports value, which is why we need to do the complement. That way, we verify that the UI effectively reflects the sports selected in the activity, (click on all sports -> the resulting sports are the complement of the initially selected sports). Hope that makes sense, don't hesitate if you need more details.

app/src/androidTest/java/com/github/sdpcoachme/profile/SelectSportsActivityTest.kt
```diff
@@ -64,10 +54,40 @@ open class SelectSportsActivityTest {
...
+            composeTestRule.onNodeWithTag(CANCEL).assertIsDisplayed()
             ROW_TEXT_LIST.forEach {
-                composeTestRule.onNodeWithTag(it.ICON).assertDoesNotExist()
+                composeTestRule.onNodeWithTag(it.ICON, useUnmergedTree = true).assertDoesNotExist()
+            }
+        }
+    }
+
+    @Test
+    fun parametersDisplayedCorrectly() {
```
Thymian18, line 69: Maybe here as well, you could choose a more meaningful name such as "parametersDisplayedCorrectlyWhenSportsWereChosenAlready" so that we can already guess from the title what the test checks.

app/src/main/java/com/github/sdpcoachme/profile/SelectSportsActivity.kt
```diff
@@ -1,87 +1,203 @@
...
+         *
+         * @param context The context of the caller activity.
+         * @param initialValue The initial selected sports to display.
+         * @param title The title displayed in the app bar.
+         * @return An Intent that can be used to launch this activity.
+         */
+        fun getIntent(
+            context: Context,
+            title: String? = null,
+            initialValue: List<Sports>? = null
```
Thymian18, line 91: Maybe, "initialSports" would be a bit clearer.

Thymian18 (changes requested):
There are some suggestions, mostly regarding the naming of functions and parameters. But overall very good work. You got rid of some code smells and implemented a lot of error handling, which will make our app more robust.

app/src/androidTest/java/com/github/sdpcoachme/profile/ProfileActivityTest.kt
Thymian18: Okay, makes sense:)

brygotti:
Test names were updated in latest commit.

PR #114, opened 2023-04-24
Email display fix

Ayskin:
This pull request (2 last commits) fixes some small bug where the wrong email was displayed in the Dashboard header on viewing a coach profile.

Some small refactoring has also been made as well as layout adapting to dark mode on map markers.


Luca-Engel (changes requested):
Great job with the bugfix and the refactoring! The code is very readable and the tests are cleary labeled.

I still found a branch that currently remains uncovered by the tests:
For the "email.isEmpty()" inside the DrawerHeader of the Dashboard, only the case where the email is not empty is tested. Once the second branch is covered, we can merge. Nice job!

Ayskin:
I've added the test for this edge case. In fact, this branch cannot be reached from an activity context as the check is already done before building our composables. So I've decided to refactor the DashboardTest class to make the UI-specific tests activity-independent. Thank you for the catch !!!

Luca-Engel (approved):
Nice refactoring, LGTM!

PR #115, opened 2023-04-25
Message caching

Luca-Engel:
#### This pull request implements and tests the caching database methods for the in-app messaging part.
 
As the stale data prevention will be done in a separate task in this sprint, this has not yet been implemented so that it can be done in a unified way.
 
#### Testing:
To make tests clearer to understand and to test the caching functionality more easily, the MockDatabase has been refactored to be an open class so that classes can extend it. This allows us to overwrite certain methods of the MockDatabase (e.g., to allow us to see how many times they are called and to make sure that they are not called when the result is already cached).
Therefore, most tests in this pull request have a locally created database extending MockDatabase and overriding the method(s) needed for the caching.
 
Also, the FireDatabase's sendMessage has been adapted to place the chat id into the database as well. Before, this was just an empty string in the database.


app/src/androidTest/java/com/github/sdpcoachme/database/CachingDatabaseTest.kt
```diff
@@ -96,12 +101,272 @@ class CachingDatabaseTest {
...
+                            }
+                    }
+            }.get(5, SECONDS)
 
-    val exampleEmail = "example@email.com"
+        assertTrue(isCorrect)
+    }
+
+    @Test
+    fun sendForNotCachedChatDoesNotCacheTheChat() {
```
Thymian18, line 290: The name here should rather be "markMessageAsReadForNotCachedChatDoesNotCacheTheChat", no? Otherwise, it is a bit confusing as you already have a test called "sendingMessageForNotCachedChatDoesNotCacheTheChat".

app/src/androidTest/java/com/github/sdpcoachme/database/CachingDatabaseTest.kt
```diff
@@ -96,12 +101,272 @@ class CachingDatabaseTest {
...
+                    }.exceptionally {
+                        false
+                    }
+            }.get(5, SECONDS)
+
+        assertTrue(isCorrect)
+    }
+
+    @Test
+    fun addChatListenerPropagatesToWrappedDatabase() {
```
Thymian18, line 312: Very smart test!

Thymian18 (changes requested):
Great idea to make the MockDatabase an open class. That expands the possibilities of proper testing a lot. I can use that for my tests too. 
All in all, everything implemented in a clear and clean way. There is only one comment I made about a renaming suggestion. Other than that, it is ready to be merged.

app/src/androidTest/java/com/github/sdpcoachme/database/CachingDatabaseTest.kt
Luca-Engel: Great catch, I will do that!

Luca-Engel:
Thank you for the name catch. I have updated the name of the test as requested in the review and also merged the updated main into the branch.

PR #116, opened 2023-04-26
Switch to firebase database linked to the google cloud project

brygotti:
This allows to use the firebase database linked to our google cloud project so we can share the billing information between firebase and other google cloud related services. It is a simple update of a settings file.

Note: In this file, the Firebase API key is available in plain text, which is not ideal. However, it has been restricted to only work with Android applications that have a known SHA-1 (that is, a SHA-1 from the development team or SDP staff). That way, it should be safe from internet bots. The API key used for Google Maps has also been restricted in the same way (which fortunately, doesn't seem to affect the CI!).


Ayskin (approved):
LGTM ! Only need to update this branch to latest main version.

PR #117, opened 2023-04-27
Schedule cache

Thymian18:
# Caching with prefetching
This PR will add caching to the schedule. 
This version always caches the currently displayed week's events as well as all the events in the upcoming 4 weeks and the 4 weeks before (compared to the currently shown week). 
As soon as the currently shown week reaches the last cached week, so the boundary on the left or right side, the cache prefetches events so that only the relevant events are in the cache. 

Additionally, I added tests to the caching DB, which I wrote in a similar manner as the ones Luca wrote for the Chats.


Ayskin (approved):
Really nice cache implementation as well as for the testing part. 

I've notived that the date display changes on March : 
Screenshot from 2023-04-27 17-31-29
But as it is independent from this caching implementation, it can be addressed later.
Also I've seen that the right arrow UI is now of fixed size :) !!

PR #119, opened 2023-05-01
Create README.md

Luca-Engel:
This pull request creates a simple readme for our project. We can update and adapt this file as we continue to add features.


Thymian18 (approved):
Very good idea to have a readme. Thank you for writing it!

PR #121, opened 2023-05-02
Schedule addevents

Thymian18:
# Add custom events
This PR will add the possibility for users to add their own events to the schedule. 
In the schedule, one can see an icon button on the bottom right. When clicking that button, a new activity will be launched. In that activity, users can choose the event title, start, end, color, and description. Once the event is confirmed, it will be added to the database and will be displayed in the schedule.

For picking the date, time, and color, an existing library was used. This allowed me to reuse existing pickers that work. The downside of using that library is that all those pickers come without modifier parameters. So, it is not possible to assign a testtag to elements that are internal to a picker. That made it a bit harder to test everything. Some cases, I was not able to test. 
Anyway, considering the big amount of code that this PR will add, the missing test cases won't decrease code coverage (or at least not much).

### DIFF
--- a/app/src/main/java/com/github/sdpcoachme/database/CachingDatabase.kt
+++ b/app/src/main/java/com/github/sdpcoachme/database/CachingDatabase.kt
@@ -74,7 +74,6 @@ class CachingDatabase(private val wrappedDatabase: Database) : Database {
     override fun getSchedule(currentWeekMonday: LocalDate): CompletableFuture<Schedule> {
         val email = wrappedDatabase.getCurrentEmail()
         currentShownMonday = currentWeekMonday
-        println("current week monday: $currentWeekMonday")
 
         if (!cachedSchedules.containsKey(email)) {  // If no cached schedule for that account, we fetch the schedule from the db
             return wrappedDatabase.getSchedule(currentWeekMonday).thenApply { schedule ->
--- a/app/src/main/java/com/github/sdpcoachme/database/CachingDatabase.kt
+++ b/app/src/main/java/com/github/sdpcoachme/database/CachingDatabase.kt
@@ -73,6 +73,7 @@ class CachingDatabase(private val wrappedDatabase: Database) : Database {
     override fun getSchedule(currentWeekMonday: LocalDate): CompletableFuture<Schedule> {
         val email = wrappedDatabase.getCurrentEmail()
         currentShownMonday = currentWeekMonday
+        println("current week monday: $currentWeekMonday")
 
         if (!cachedSchedules.containsKey(email)) {  // If no cached schedule for that account, we fetch the schedule from the db
             return wrappedDatabase.getSchedule(currentWeekMonday).thenApply { schedule ->
--- a/app/src/main/java/com/github/sdpcoachme/database/CachingDatabase.kt
+++ b/app/src/main/java/com/github/sdpcoachme/database/CachingDatabase.kt
@@ -54,12 +54,12 @@ class CachingDatabase(private val wrappedDatabase: Database) : Database {
     }
 
     // Note: to efficiently use caching, we do not use the wrappedDatabase's addEventsToUser method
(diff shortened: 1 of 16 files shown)


app/src/main/java/com/github/sdpcoachme/database/CachingDatabase.kt
```diff
@@ -74,6 +74,7 @@ class CachingDatabase(private val wrappedDatabase: Database) : Database {
     override fun getSchedule(currentWeekMonday: LocalDate): CompletableFuture<Schedule> {
         val email = wrappedDatabase.getCurrentEmail()
         currentShownMonday = currentWeekMonday
+        println("current week monday: $currentWeekMonday")
```
Ayskin, line 77: This print can be removed

app/src/main/java/com/github/sdpcoachme/database/Database.kt
```diff
@@ -70,7 +70,7 @@ interface Database {
      * @param currentWeekMonday The monday of the current week
      * @return A future with currently stored schedule that will complete when the events have been added.
      */
-    fun addEvents(events: List<Event>, currentWeekMonday: LocalDate): CompletableFuture<Schedule>
+    fun addEvent(event: Event, currentWeekMonday: LocalDate): CompletableFuture<Schedule>
```
Ayskin, line 73: The documentation of this function can be updated.

app/src/main/java/com/github/sdpcoachme/schedule/ScheduleActivity.kt
```diff
@@ -176,34 +171,53 @@ fun Schedule(
...
-            events = eventsToShow.filter {event ->
-                val eventDate = LocalDateTime.parse(event.start).toLocalDate()
-                eventDate >= shownWeekMonday && eventDate < shownWeekMonday.plusWeeks(1)
-            },
-            shownMonday = shownWeekMonday,
-            dayWidth = dayWidth,
+            println("Get shown events")
+            // filter events to only show events in the current week
+            val eventsToShow = EventOps.eventsToWrappedEvents(events)
+            println("Events to show: $eventsToShow")
```
Ayskin, line 192: Two prints left here (192 & 189)

app/src/test/java/com/github/sdpcoachme/ExampleUnitTest.kt
```diff
@@ -14,4 +16,16 @@ class ExampleUnitTest {
     fun addition_isCorrect() {
         assertEquals(4, 2 + 2)
     }
+
+    @Test
+    fun colorTest() {
```
Ayskin, line 21: Does this test needs to be merged into production ?

app/src/main/java/com/github/sdpcoachme/schedule/CreateEventActivity.kt
```diff
@@ -0,0 +1,609 @@
...
+                            tint = MaterialTheme.colors.onPrimary,
+                            modifier = Modifier.testTag(CreateEventActivity.TestTags.Icons.CANCEL_ICON),
+                        )
+                    }
+                },
+                actions = {
+                    IconButton(
+                        onClick = {
+                            EventOps.addEvent(event, database).thenAccept {
+                                goBackToScheduleActivity()
```
Ayskin, line 221: Here, before saving the new creating event, we could add a check that entered parameters are valid, ie startDate < endDate. If not we can just show a popup "invalid parameters" and wait for the user to correct them.

Ayskin (changes requested):
Great new functionnality that fulfills the schedule feature. Very nice test classes and great refactoring ! You can have a look at the comments but I just mainly suggest to add some input checks in the CreateEventActivity when the user tries to save what he has entered (see related comment).

app/src/test/java/com/github/sdpcoachme/ExampleUnitTest.kt
Thymian18: No, you are right, I deleted my test as it was mainly for me.

Ayskin (approved):

PR #122, opened 2023-05-03
Device location

Ayskin:
This branch refactors the location handling implementation. 

This includes new features :
- the user can now use the application without the device location activated.
- the user address is set as default location. The user is alerted via a popup.
- the user can now enable the device location if disabled on application launch.

The code includes the following changes :
- A refactoring of "location" containing names that were assigned to address handling. There is now a clear distinction between the user location and the user address / address of a coach. This requires the database to store values serialized following this new naming.
- All location-related work done in the application has been moved to a `LocationProvider`, held in the application-wide context (instead of the previous `lastLocation` state). This `LocationProvider` is now reachable from any activity and can be freely initialized as needed.
- A new `MockLocationProvider` to handle the state-flow of the application `LocationProvider` in tests.
- Fix merge conflicts due to this new location handler implementation.

The choice to move the location "handler" to a `LocationProvider` interface out of the `MapActivity` has been partly made for testing purposes. The implementation inside the instantiation class `FusedLocationProvider` is roughly the same as before but includes now the setting check. Official documentation often provides deprecated code, so work-arounds have been found.


Ayskin:
Overall coverage is affected by the `FusedLocationProvider` missing coverage. Tests of `FusedLocationProviderTest` work locally but not on the CI. Should be addressed later on.

app/src/main/java/com/github/sdpcoachme/location/provider/FusedLocationProvider.kt
```diff
@@ -0,0 +1,184 @@
...
+                        // lastLocation task would not complete successfully nor could I find any
+                        // suitable implementation.
+                        getDeviceLocation(delay + 1)
+                    }
+                }
+            }
+        } catch (e: SecurityException) {
+            error("getDeviceLocation was called without correct permissions : ${e.message}")
+        }
+    }
```
lindsaybordier, line 131: I see the issue here... Here is the solution I thought about:

- Create a future eg. f.
- Lauch a coroutine (Globalscope.launch...) with a while loop testing if we successfully get the location. Complete the future once successful.
- Have the rest of the operations done asynchronously for example with thenApply

Hope this can help.

app/src/main/java/com/github/sdpcoachme/location/provider/FusedLocationProvider.kt
```diff
@@ -0,0 +1,184 @@
...
+     *
+     * Notice that the fusedLocationProviderClient.lastLocation task is null if the location is
+     * disabled on the device (also even if the last location was previously retrieved because
+     * disabling the device location clears the cache).
+     */
+    @SuppressLint("MissingPermission") //permission is checked before the call
+    private fun getDeviceLocation(delay: Long) {
+        // getDeviceLocation should not be called more than MAX_LOCATION_DELAY time
+        if (delay >= DELAY) {
+            error("getDeviceLocation has reached its max recursive delay")
```
lindsaybordier, line 112: I have noticed that the termination is based on this delay, could you explain how it was chosen?
Also, the documentation could be updated above.

lindsaybordier (changes requested):
Thanks for the great refactoring! 
Really nice moving all the logic related to the location in the FusedLocationProvider. The MapActivity is now very clean and super easy to understand and maintain.

Great job for the modularisation as well as the mocking for the testing.

Please take a look at the few comments. I also proposed a solution for the issue of spinning in FusedLocationProvider.

app/src/main/java/com/github/sdpcoachme/location/provider/FusedLocationProvider.kt
Ayskin: The DELAY has been set to 5000 iterations after several local tests. `getDeviceLocation(...)` should normally not be called over around 3000 times. If it happens, delay should be adjusted or another implementation found.

app/src/main/java/com/github/sdpcoachme/location/provider/FusedLocationProvider.kt
Ayskin: Yes this may be a way to implement this. The problem of the while loop is that we cannot catch the completion of the setting enabling so in the end, it would be an arbitrary TimeOut. In fact, this would replace the same behavior as using a manually set maximum recurive calls DELAY.

lindsaybordier (approved):

PR #123, opened 2023-05-05
Filter by sport

Ayskin:
This PR adds a button in nearby coaches that enables the user to apply a filter by sports in the displayed list of coaches. Only the coaches having the selected sports as favorites are displayed when the filter is applied, allowing the user to quickly find the best coaching following his needs.

Conflicts with the LocationProvider were fixed and tests were added to CoachesListActivityTest.

### DIFF
--- a/app/src/main/java/com/github/sdpcoachme/profile/CoachesListActivity.kt
+++ b/app/src/main/java/com/github/sdpcoachme/profile/CoachesListActivity.kt
@@ -136,7 +136,8 @@ class CoachesListActivity : ComponentActivity() {
             LazyColumn {
                 items(listOfCoaches) {user ->
                     // Filtering should not influence the coaches list in contacts view
-                    if (!Collections.disjoint(user.sports, sportsFilter.value)) {
+                    // We still show user with no favourite sports, especially for testing purposes
+                    if (user.sports.isEmpty() || !Collections.disjoint(user.sports, sportsFilter.value)) {
                         UserInfoListItem(user, isViewingContacts)
                     }
                 }
--- a/app/src/main/java/com/github/sdpcoachme/profile/CoachesListActivity.kt
+++ b/app/src/main/java/com/github/sdpcoachme/profile/CoachesListActivity.kt
@@ -63,9 +63,12 @@ class CoachesListActivity : ComponentActivity() {
 
         val isViewingContacts = intent.getBooleanExtra("isViewingContacts", false)
         val database = (application as CoachMeApplication).database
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/github/sdpcoachme/profile/CoachesListActivity.kt
```diff
@@ -11,45 +11,69 @@ import androidx.compose.foundation.layout.*
...
+                const val FILTER = "filterSearch"
+            }
+        }
+    }
+
     // Allows to notice testing framework that the activity is ready
     var stateLoading = CompletableFuture<Void>()
+
+    // Observable state of the current sports used to filter the coaches list
+    private lateinit var sportsFilter: MutableState<List<Sports>>
```
brygotti, line 58: This mutable state can be moved in the composable CoachesList for better readability of the code! The syntax would be `var sportsFilter by remember { mutableStateOf(Sports.values().toList()) }` and then setting and getting can be done via standard kotlin variable syntax.

app/src/main/java/com/github/sdpcoachme/profile/CoachesListActivity.kt
```diff
@@ -11,45 +11,69 @@ import androidx.compose.foundation.layout.*
...
         val isViewingContacts = intent.getBooleanExtra("isViewingContacts", false)
         val database = (application as CoachMeApplication).database
-        val lastLocation = (application as CoachMeApplication).locationProvider.getLastLocation()
-        val userLatLng = lastLocation.value?: CAMPUS
         val email = database.getCurrentEmail()
 
+        val locationProvider = (application as CoachMeApplication).locationProvider
+        // Here we don't need the UserInfo
+        locationProvider.init(this, CompletableFuture.completedFuture(null))
+        val userLatLng = locationProvider.getLastLocation().value?: CAMPUS
```
brygotti, line 71: Adding a small comment explaining when getLastLocation can return null could be helpful to better understand this small bit of code.

brygotti (changes requested):
Very good job! Please have a look at my comments for the minor suggestions, but overall, the code is super clean. Thanks for refactoring small details like Intents.init() / release() in tests, this allows to keep the code base in good shape.

app/src/main/java/com/github/sdpcoachme/profile/CoachesListActivity.kt
Ayskin: Yes, this is better. Thanks for the tip !

brygotti (approved):

PR #145, opened 2023-05-15
Data coherence persistence

lindsaybordier:
This pull request will provide the new Caching Store architecture.
The whole code base had to be adapted, including tests.

Here are the features:

- The Caching Store is a global object accessible anywhere in the app that holds the state of the app.
- It wraps a Database and a Datastore object to communicate both with the database and local device storage.
- It implements caching and stale data prevention for all data in the app. 
- Offline mode now works for accessing Nearby Coaches, Contacts, Messages, Profile etc... even the app wall fully closed.
- If the user already logged in and closes the app entirely, he does not have to login again where re-opening the app.
- Same kinds of functionality holds for messages and schedule.
- The storage on the device is made by serialising using Gson and storing using a datastore instance. There needs to be only one datastore instance per file, otherwise the app crashes (see doc).
- Internet connectivity is checked when launching the app (in Caching Store) to decide wether to update all caches or keep previously stored data (on device disk) in offline mode.

Important things to note: 

- The current user email is now provided as a future as it is retrieved from the local device storage in an asynchronous manner during the app launch (discussed with team). It needs to be used asynchronously.
- Especially in tests, it is important to wail for the store's retrieveData future to complete before setting the email to avoir races as the locally stored data is fetched asynchronously.
- Some problems were encountered to make all tests pass again, especially some tests were passing locally but not on the CI, everything works fine now.


app/src/androidTest/java/com/github/sdpcoachme/CoachMeTestApplication.kt
```diff
@@ -32,13 +38,26 @@ class CoachMeTestApplication : CoachMeApplication() {
...
          */
-        database = MockDatabase()
+        store = CachingStore(MockDatabase(), dataStore, this)
         locationProvider = MockLocationProvider()
-
         // Might be necessary to initialize Places SDK, but for now, we don't need it.
     }
     override val authenticator: Authenticator = MockAuthenticator()
 
+    override fun onTerminate() {
```
Thymian18, line 47: The naming of superOnTerminate is a bit confusing, maybe you could add a small comment.

app/src/androidTest/java/com/github/sdpcoachme/auth/LoginActivityTest.kt
```diff
@@ -5,42 +5,42 @@ import androidx.compose.ui.test.*
...
 open class LoginActivityTest {
     private val launchTimeout = 5000L
     private lateinit var device: UiDevice
-    private lateinit var signedOutInfoText: String
-    private lateinit var deleteInfoText: String
 
-    @Before
-    open fun startLoginActivityFromHomeScreen() {
+    @Test
+    fun startFromHomeScreenLaunchesLoginActivity() {
```
Thymian18, line 41: Good renaming

app/src/androidTest/java/com/github/sdpcoachme/auth/SignupActivityTest.kt
```diff
@@ -142,26 +134,45 @@ open class SignupActivityTest {
 
         // Important note: this get method was used instead of onTimeout due to onTimeout not
         // being found when running tests on Cirrus CI even with java version changed in build.gradle
-        val retrievedUser = database.getUser(user.email).get(10, SECONDS)
+        /*val retrievedUser = database.getUser(user.email).get(10, SECONDS)
```
Thymian18, line 137: Remove those comments please.

app/src/androidTest/java/com/github/sdpcoachme/database/MockDatabase.kt
Thymian18, line 1: The existing functions were correctly adapted to the new use of emails.

app/src/androidTest/java/com/github/sdpcoachme/database/CachingStoreTest.kt
```diff
@@ -14,171 +21,189 @@ import com.github.sdpcoachme.data.messaging.Message
...
-        val wrappedDatabase = MockDatabase()
-        val cachingDatabase = CachingDatabase(wrappedDatabase)
-        val retrievedUser = cachingDatabase.updateUser(willSmithUser)
-            .thenCompose { cachingDatabase.getUser(willSmithUser.email) }
+        val retrievedUser = cachingStore.updateUser(willSmithUser)
+            .thenCompose { cachingStore.getUser(willSmithUser.email) }
             .get(5, SECONDS)
-        assertTrue(cachingDatabase.isCached(willSmithUser.email))
-        assertTrue(cachingDatabase.userExists(willSmithUser.email).get(1, SECONDS))
+        assertTrue(cachingStore.isCached(willSmithUser.email))
```
Thymian18, line 83: This test does not really test if updateUser caches the user since you always call getUser first and getUser is supposed to cache the user already. So, this would pass even if updateUser did not cache the user since it is already cached at that point. You can have a class that extends the MockDatabase to overwrite updateUser, just like it is done below.

app/src/androidTest/java/com/github/sdpcoachme/database/CachingStoreTest.kt
```diff
@@ -14,171 +21,189 @@ import com.github.sdpcoachme.data.messaging.Message
...
+        cachingStore = CachingStore(wrappedDatabase,
+            ApplicationProvider.getApplicationContext<Context>().dataStoreTest,
+            ApplicationProvider.getApplicationContext()
+        )
+        cachingStore.retrieveData.get(1, SECONDS)
+        val updatedUser = cachingStore.getUser(exampleEmail)
             .thenCompose {
-                cachingDatabase.updateUser(defaultUser) }
-            .thenCompose { cachingDatabase.getUser(exampleEmail) }
+                cachingStore.updateUser(defaultUser) }
```
Thymian18, line 119: Here, we have the same issue as described in the comment above.

app/src/androidTest/java/com/github/sdpcoachme/database/CachingStoreTest.kt
```diff
@@ -585,36 +646,49 @@ class CachingDatabaseTest {
...
 
-    private val defaultUser = UserInfo(
-        "John",
-        "Doe",
-        exampleEmail,
-        "1234567890",
-        LAUSANNE,
-        false
-    )
+    companion object {
```
Thymian18, line 652: To have all the test users in one place, you can just use the existing test users of UserInfoSamples. (change in next PR)

app/src/androidTest/java/com/github/sdpcoachme/messaging/ChatActivityTest.kt
```diff
@@ -179,30 +178,6 @@ class ChatActivityTest {
 
             composeTestRule.onNodeWithText("${toUser.firstName} ${toUser.lastName}")
                 .assertIsDisplayed()
-            Intents.release()
-        }
-    }
-
-    @Test
-    fun chatListenerAddedAtStartUp() {
```
Thymian18, line 187: Why is that test removed? This should not be the case I think.

app/src/androidTest/java/com/github/sdpcoachme/messaging/ChatActivityTest.kt
```diff
@@ -179,30 +178,6 @@ class ChatActivityTest {
...
-        val mockDB = database as MockDatabase
-        assertThat(mockDB.numberOfAddChatListenerCalls(), `is`(0))
-
-        ActivityScenario.launch<ChatActivity>(defaultIntent).use {
-            assertThat(mockDB.numberOfAddChatListenerCalls(), `is`(1))
-        }
-    }
-
-    @Test
-    fun pressingBackButtonRemovesChatListener() {
```
Thymian18, line 197: Why is that test removed? This should not be the case I think.

app/src/main/java/com/github/sdpcoachme/CoachMeApplication.kt
```diff
@@ -16,24 +16,41 @@ import com.google.android.libraries.places.api.Places
...
         // Initialize Places SDK
         Places.initialize(applicationContext, BuildConfig.MAPS_API_KEY)
     }
 
+    override fun onTerminate() {
+        super.onTerminate()
+        store.storeLocalData()
+    }
+
+    fun superOnTerminate() {
```
Thymian18, line 50: As mentioned in CoachMeTestApplication, this naming is not a problem but a bit confusing, maybe add an explanatory comment.

app/src/main/java/com/github/sdpcoachme/auth/LoginActivity.kt
```diff
@@ -17,94 +17,101 @@ import androidx.compose.foundation.layout.padding
...
-                .putExtra("toUserEmail", pushNotificationIntent.getStringExtra("sender"))
-
-            // If the user is logged in, then we can open the chat activity directly without letting the user log in,
-            // otherwise we wait for the login process to complete and then open the chat activity
-            if (database.getCurrentEmail().isNotEmpty()) {
-                startActivity(userExistsIntent)
+        store =  (application as CoachMeApplication).store
+
+        store.isLoggedIn().thenAccept { loggedIn ->
+            if (loggedIn) {
```
Thymian18, line 102: This is written in a much clearer way before, very good.

app/src/main/java/com/github/sdpcoachme/database/CachingStore.kt
Thymian18, line 1: This class is going to be adapted a lot during the PRs of the other team members. It seems to have all the functions that were in the caching database however. The idea of having a caching store class was discussed during a standup meeting and you implemented it as discussed, it seems.

app/src/main/java/com/github/sdpcoachme/database/FireDatabase.kt
```diff
@@ -37,49 +36,27 @@ class FireDatabase(databaseReference: DatabaseReference) : Database {
         return getChild(accounts, userID).thenApply { it.getValue(UserInfo::class.java) }
     }
 
-    override fun getAllUsers(): CompletableFuture<List<UserInfo>> {
```
Thymian18, line 40: Why is that removed?

app/src/main/java/com/github/sdpcoachme/messaging/InAppNotifier.kt
```diff
@@ -59,34 +59,36 @@ class InAppNotifier(val context: Context, val database: Database) {
     private fun sendMessagingNotification(notificationTitle: String, notificationBody: String, sender: String) {
 
         // The more info we receive, the more we can customize the notification's behaviour (up until the chat itself)
-        val intent: Intent
-        when {
-            database.getCurrentEmail().isEmpty() -> {
-                intent = Intent(context, LoginActivity::class.java)
+        store.isLoggedIn().thenAccept { isLoggedIn ->
```
Thymian18, line 62: Good implementation of new login.

Thymian18 (changes requested):
This was a big task and needed a lot of refactoring. The fact that the new caching system affected pretty much all classes made it difficult to not forget anything. Anyway, the app (functionality- and featurewise) seems to be on the version as we had it in the end of last sprint, which is really good for the further PRs in terms of merging. 

A small suggestion for the next PR: Instead of calling .get() on every future in the tests and then having the asserts outside, I think it would be good to chain the calls. This way, there is only one .get() call at the end of the whole chain. This will improve maintainability, since in case one wants to test things with delay, you only have to adapt the waiting time in that one .get() call. 

I made some in-line comments, please resolve them before we merge.

lindsaybordier:
Changes were made accordingly, thanks for the review!

I think that the two deleted tests were deleted by someone else that coded on my branch to fix some tests, but I might be wrong. Anyway we can put those tests back in the next PR as this code needs to be merged quickly for the rest of the team.

Thymian18 (approved):
Th requested changes look sufficient for that first merge so that the team can continue from here. As discussed, a second PR will be done with refinements in the code.

PR #146, opened 2023-05-16
Groupevents database prep

Thymian18:
This PR adds the necessary functions that are needed for GroupEvents to the DB related classes. The behaviour of the app won't change with this PR, it is only a preparation for the remaining PRs of this sprint.

There are some branches that I will be able to cover in the future. But since others are depending on that PR, I will increase code coverage during the PR that adds the ability to actually let a coach add a group event. 
The coverage should still be fine without those refinements. It will go down a bit because I had to comment a test that failed, related to the earlier PR of Lindsay, but we will fix this test after the standup meeting as discussed with him, so that in his second PR, the test works.


app/src/androidTest/java/com/github/sdpcoachme/data/UserAddressSamples.kt
```diff
@@ -2,37 +2,37 @@ package com.github.sdpcoachme.data
 
 class UserAddressSamples {
     companion object {
-        val PARIS = UserAddress(
+        val PARIS = Address(
```
Luca-Engel, line 5: Great renaming! This makes it clearer to use for future coding

app/src/androidTest/java/com/github/sdpcoachme/database/CachingStoreTest.kt
```diff
@@ -187,25 +202,173 @@ class CachingStoreTest {
...
+            failFuture.completeExceptionally(NoSuchElementException())
+            return failFuture
         }
+
+        override fun getSchedule(email: String, currentWeekMonday: LocalDate): CompletableFuture<Schedule> {
+            return CompletableFuture.completedFuture(schedule)
+        }
+    }
+    @Test
+    fun registerForGroupEventAddsParticipantToWrappedDatabase() {
```
Luca-Engel, line 276: Great test case! Also checking that the user was added to every event is great!

app/src/androidTest/java/com/github/sdpcoachme/database/CachingStoreTest.kt
```diff
@@ -254,6 +416,78 @@ class CachingStoreTest {
...
+                true
+            }.exceptionally {
+                false
+            }.get(5, SECONDS)
+
+        assertTrue(isCorrect)
+    }
+
+    @Test
+    fun getGroupEventFailsForUnregisteredUser() {
```
Luca-Engel, line 470: Nice job also testing this edge case!

app/src/main/java/com/github/sdpcoachme/database/FireDatabase.kt
```diff
@@ -49,12 +53,55 @@ class FireDatabase(databaseReference: DatabaseReference) : Database {
         }
     }
 
+    override fun addGroupEvent(groupEvent: GroupEvent): CompletableFuture<Void> {
+        var errorPreventionFuture = CompletableFuture<Void>()
+
+        if (groupEvent.participants.size > groupEvent.maxParticipants) {
+            errorPreventionFuture.completeExceptionally(Exception("Group event should not be full, initially"))
+        } else if (groupEvent.participants.isEmpty()) {
+            errorPreventionFuture.completeExceptionally(Exception("Group event must have at least 2 participants"))
```
Luca-Engel, line 62: Shouldn't this error rather say "... at least 1 participant"?

app/src/main/java/com/github/sdpcoachme/database/FireDatabase.kt
```diff
@@ -49,12 +53,55 @@ class FireDatabase(databaseReference: DatabaseReference) : Database {
...
+        }
+    }
+
     override fun getSchedule(email: String, currentWeekMonday: LocalDate): CompletableFuture<Schedule> {
         val id = email.replace('.', ',')
         return getChild(schedule, id).thenApply { it.getValue(Schedule::class.java)!! }
             .exceptionally { Schedule() }
     }
 
+    override fun getGroupEvent(groupEventId: String): CompletableFuture<GroupEvent> {
```
Luca-Engel, line 100: I believe, it would be best to also call replace on the groupEventId to replace "."'s by ","'s

app/src/main/java/com/github/sdpcoachme/schedule/CreateEventActivity.kt
```diff
@@ -223,11 +231,23 @@ fun NewEvent(store: CachingStore) {
                                 val toast = Toast.makeText(context, "Start date must be before end date", Toast.LENGTH_SHORT)
                                 toast.show()
                             } else {
-                                EventOps.addEvent(event, store).thenAccept {
+                                EventOps.addEvent(event, store)/*.thenCompose {
+                                    val organiser = database.getCurrentEmail().replace('.', ',')
```
Luca-Engel, line 235: Will this commented out code snipped be needed in a future pr? Otherwise I would recommend to delete it.

Luca-Engel (changes requested):
Great job on this pull request! I have added some comments mainly for some tests and other minor potential improvements.

I also agree with the choice to do this pr now and add more tests later on as multiple other sprint tasks depend on the skeleton of this one.

Once the issues have been addressed, this pr will be ready to merge into main. Great job!

app/src/main/java/com/github/sdpcoachme/schedule/CreateEventActivity.kt
Thymian18: Yes, this will be used for the implementation of the next PR.

Luca-Engel (approved):
Great job, LGTM!

PR #147, opened 2023-05-17
Event group chats

Luca-Engel:
### This pull request adds the group chat feature for group events.
As the group events have not yet been pulled to main yet, this feature cannot be accessed yet for users without manually adding a group event and group chat to the database. This, however, has been done for thorough testing and works. The reason for creating a PR for this feature now and not once group event displaying and creation is implemented is to increase flexibility of the future PRs to prevent multiple PRs to be done last minute. Additionally, tests have been added and adapted for this feature.

### Specifically, the following features have been added:
- group chat creation
  - create / update group chats
  - update the participants in it
- using group chats
  - inside the group chat, the implementation is similar to the one of other messaging services such as WhatsApp: the senders name is displayed in the message, if the message has not been sent by the current user
  - when clicking on the group chat's contact field, currently a toast is shown to the user. In the future, this will open the group event and display it to the user
  - when clicking on the contact row of the group chat, 
- cloud function
  - adapted cloud function to send push notification to every participant in the chat that has not yet read the message
  - therefore, the way messages are stored in the database has been adapted
- Contact listing
  - the group chats are also listed in the contacts of the user and can be accessed through the "messaging" section
  - for every listed contact, the latest message is shown alongside the user who sent the message
  - for a group chat, the name of the group chat is shown instead of the recipient's name like with the normal chats

### Note on testing
Also for this pull request, code had to be written for the cloud function that is not covered by tests as in the last sprint. Instead, it was tested by running it on the local emulator and again after deploying it.
Lastly, the FireDatabase functions cannot be tested. Nevertheless, the test coverage remains almost the same.


app/src/androidTest/java/com/github/sdpcoachme/database/CachingStoreTest.kt
```diff
@@ -592,11 +606,200 @@ class CachingStoreTest {
...
+            ApplicationProvider.getApplicationContext()
+        )
+        cachingStore.setCurrentEmail(exampleEmail)
+
+        val chatId = "chatId"
+        val participants = listOf("user1@email.com", "user2@email")
+
+        val isCorrect = cachingStore.updateChatParticipants(chatId, participants)
+            .thenCompose { // update when not cached
+                MatcherAssert.assertThat(wrappedDatabase.nbCallsToUpdateParticipants, `is`(1))
```
Ayskin, line 643: Static imports can be used here on `MatcherAssert.assertThat`.

app/src/androidTest/java/com/github/sdpcoachme/database/CachingStoreTest.kt
```diff
@@ -942,23 +1151,6 @@ class CachingStoreTest {
             false
         )
 
-        val rogerFedererUser = UserInfo(
```
Ayskin, line 945: Sadly, Roger has to leave....

app/src/androidTest/java/com/github/sdpcoachme/database/MockDatabase.kt
```diff
@@ -209,41 +212,55 @@ open class MockDatabase: Database {
...
+        } else {
+            chat = Chat.markOtherUsersMessagesAsRead(chat, email)
+        }
 
         return CompletableFuture.completedFuture(null)
     }
 
     override fun removeChatListener(chatId: String) {
-        numberOfRemovedChatListenerCalls++
+        // Not necessary
```
Ayskin, line 257: Should we keep this function then ?

app/src/androidTest/java/com/github/sdpcoachme/profile/ContactsListActivityTest.kt
```diff
@@ -0,0 +1,169 @@
...
+            )
+        )
+    }
+
+    @Test
+    fun dashboardHasRightTitleOnContactsList() {
+        startActivity()
+        val title = (InstrumentationRegistry.getInstrumentation()
+            .targetContext.applicationContext as CoachMeApplication).getString(R.string.contacts)
+        composeTestRule.onNodeWithTag(Dashboard.TestTags.BAR_TITLE).assertExists()
```
Ayskin, line 148: Static imports for testTag can be used here

Ayskin (approved):
Tremendous work done here. Nice documentation of the public functions and nice modularization of the tests. No main changes to request. 

I'll suggest adding a main comment to the index.ts file to better explicit its behavior.

app/src/androidTest/java/com/github/sdpcoachme/database/MockDatabase.kt
Luca-Engel: As MockDatabase implements Database, we need to also have an implementation for the methods so we cannot remove them

PR #148, opened 2023-05-18
Groupevents creation

Thymian18:
# Creating group events

This PR adds the possibility for a coach to create a new group event.
A group event, besides the attributes of a normal event, includes the following attributes:

- organiser: the coach that is responsible for the group event
- maxParticipants: the number of maximum participants for the event; when reached, one cannot register anymore
- participants: list of emails of participants
- groupEventId: in the format "@@event{organiser}{start}"

Additionally, every normal event (and therefore also group events) take a sport and a location as argument.

## What changes for non-coaches
Non-coaches still directly get to the activity to create a private event when clicking on the +button in schedule. 

## What changes for coaches
A coach can now chose between creating a private event or a group event when clicking on the +button in schedule. 
When chosing to create a private event, the same event creation form will show up as for every non-coach. When choosing to create a group event, the form will have the additional fields that correspond to a group event (described above).
For now, a coach is also registered to the event once the event gets created. This simplifies things for now.

For every user that is registered for a group event (including coaches), the group event will also be displayed in the schedule.

### DIFF
--- a/app/src/androidTest/java/com/github/sdpcoachme/schedule/ScheduleActivityTest.kt
+++ b/app/src/androidTest/java/com/github/sdpcoachme/schedule/ScheduleActivityTest.kt
@@ -6,15 +6,21 @@ import androidx.compose.ui.test.junit4.createEmptyComposeRule
 import androidx.test.core.app.ActivityScenario
 import androidx.test.core.app.ApplicationProvider
 import androidx.test.espresso.intent.Intents
+import androidx.test.espresso.intent.matcher.IntentMatchers.hasComponent
+import androidx.test.espresso.intent.matcher.IntentMatchers.hasExtra
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import com.github.sdpcoachme.CoachMeApplication
 import com.github.sdpcoachme.CoachMeTestApplication
+import com.github.sdpcoachme.data.UserInfoSamples
+import com.github.sdpcoachme.data.schedule.EventType
 import com.github.sdpcoachme.data.schedule.ShownEvent
 import com.github.sdpcoachme.database.CachingStore
-import com.github.sdpcoachme.database.MockDatabase
 import com.github.sdpcoachme.errorhandling.IntentExtrasErrorHandlerActivity.TestTags.Buttons.Companion.GO_TO_LOGIN_BUTTON
 import com.github.sdpcoachme.errorhandling.IntentExtrasErrorHandlerActivity.TestTags.TextFields.Companion.ERROR_MESSAGE_FIELD
 import com.github.sdpcoachme.location.MapActivity
+import com.github.sdpcoachme.schedule.ScheduleActivity.TestTags.Buttons.Companion.ADD_EVENT_BUTTON
+import com.github.sdpcoachme.schedule.ScheduleActivity.TestTags.Buttons.Companion.ADD_GROUP_EVENT_BUTTON
+import com.github.sdpcoachme.schedule.ScheduleActivity.TestTags.Buttons.Companion.ADD_PRIVATE_EVENT_BUTTON
 import com.github.sdpcoachme.schedule.ScheduleActivity.TestTags.Buttons.Companion.LEFT_ARROW_BUTTON
 import com.github.sdpcoachme.schedule.ScheduleActivity.TestTags.Buttons.Companion.RIGHT_ARROW_BUTTON
 import com.github.sdpcoachme.schedule.ScheduleActivity.TestTags.Companion.BASIC_SCHEDULE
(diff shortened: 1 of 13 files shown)


app/src/main/java/com/github/sdpcoachme/database/CachingStore.kt
```diff
@@ -308,7 +310,7 @@ class CachingStore(private val wrappedDatabase: Database,
     /**
      * Gets the schedule for the current user
      * @param currentWeekMonday the monday of the current week
-     * @return a completable future that completes when the schedule has been retrieved
+     * @return a completable future that completes when the schedule has been retrieved (containing the newly cached schedule)
```
lindsaybordier, line 313: Good clarification

app/src/main/java/com/github/sdpcoachme/database/Database.kt
```diff
@@ -68,7 +68,7 @@ interface Database {
      * @param groupEventId The id of the group event to add the participant to
      * @return A future that will complete when the participant has been added.
```
lindsaybordier, line 69: Would be great to update the documentation here

app/src/main/java/com/github/sdpcoachme/database/FireDatabase.kt
```diff
@@ -59,33 +59,34 @@ class FireDatabase(databaseReference: DatabaseReference) : Database {
...
             errorPreventionFuture.completeExceptionally(Exception("Group event should not be full, initially"))
-        } else if (groupEvent.participants.isEmpty()) {
-            errorPreventionFuture.completeExceptionally(Exception("Group event must have at least 1 participants"))
         } else if (LocalDateTime.parse(groupEvent.event.start).isBefore(LocalDateTime.now())) {
             errorPreventionFuture.completeExceptionally(Exception("Group event cannot be in the past"))
         } else {
-            errorPreventionFuture = setChild(groupEvents, groupEvent.groupEventId, groupEvent).thenCompose {
-                registerForGroupEvent(groupEvent.organiser, groupEvent.groupEventId)
-            }
+            errorPreventionFuture = setChild(groupEvents, groupEvent.groupEventId, groupEvent)
```
lindsaybordier, line 65: Great job on moving that piece of code to the CachingStore. Follows great logic: FireDatabase should provide easy database access methods without unpredictable side effects, therefore it is better to chain those actions in CachingStore.

app/src/main/java/com/github/sdpcoachme/schedule/CreateEventActivity.kt
```diff
@@ -151,487 +169,646 @@ class CreateEventActivity : ComponentActivity() {
 
     private lateinit var store: CachingStore
     private lateinit var email: String
+    private lateinit var addressAutocompleteHandler: AddressAutocompleteHandler
+    private lateinit var selectSportsHandler: (Intent) -> CompletableFuture<List<Sports>>
+
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         store = (application as CoachMeApplication).store
         email = store.getCurrentEmail().get(1000, TimeUnit.MILLISECONDS)
```
lindsaybordier, line 178: The email should be handled asynchronously and should not block the UI thread. I see that you did not modify this code, but it should be better handled just like in other activities.

app/src/androidTest/java/com/github/sdpcoachme/schedule/ScheduleActivityTest.kt
```diff
@@ -263,7 +270,7 @@ class ScheduleActivityTest {
         val previousWeekEvent = EventOps.getPreviousWeekEvent()
 
         store.addEvent(previousWeekEvent, currentMonday).thenRun {
-            store.setCurrentEmail(defaultEmail)
+            store.setCurrentEmail(coachEmail)
```
lindsaybordier, line 273: Be careful, setCurrentEmail returns a future. Would be better to apply a get(1, SECONDS) or use .thenApply before executing the rest of the code

lindsaybordier (changes requested):
This is overall great code and a significant amount of work. Thanks for the effort :))

Please take a look at my comments on the code.

Some more general comments: 

- When running the app locally and creating a group event, if I select the field to enter the max number of participants and press backspace (keyboard or button), the activity crashes with the following error:

image

- Concerning the backend, I think it would be better to have the "groupEvents" node outside of the schedule subsection of the database - just like Messages, Accounts etc... I don't think it is great to mix this independent section in the middle of the user emails used as keys for schedules, especially for clarity.

- When running the app and creating a group event, it did not seem to appear in my schedule at the defined them. After discussion, it should not be too hard to fix.

Thanks

app/src/main/java/com/github/sdpcoachme/schedule/CreateEventActivity.kt
Thymian18: Yes actually, we don't even need the email in that activity anymore. I will take it out.

lindsaybordier (approved):

PR #149, opened 2023-05-18
Hotfix

Thymian18:
# Bugfix in case of group event creation
After having done the last PR, I discovered a bug that was to be fixed as quick as possible. When creating a group event, the coach gets redirected to the schedule. The newly added group event however was not displayed immediately however. 
Now, the group event gets displayed in the schedule. 
Also the default time was adapted to be more robust to testing.


Luca-Engel (approved):
Great job on the catch of this bug! I know that the fix took a long time so we really appreciate the extra effort you did!

PR #154, opened 2023-05-19
Weather forecast

Ayskin:
**Summary** : This branch implements the whole weather forecast pipeline and displays the forecast in the schedule.

**Design** : the structure chosen follows an MVP-like design which was deployed with the following layers :

- `WeatherApi` : makes an request to open-meteo api
- `WeatherRepository` : calls the `WeatherApi` and parses the retrieved Data
- `WeatherPresenter` : holds a repository and provies two core states; one observableWeatherState for any view to allow recomposition as the weather forecast retrieval is done asynchronously (same behavior as the locationState passed to our `MapView`); one simple WeatherState that is cached in the `CachingStore` and written to the disk.
- `WeatherView` : implementation of a WeatherView that displays a weather icon and the max/min temperature

**Api choice** : The open-meteo api has been chosen for its surprising simplicity and our application only retrieves a forecast or the upcoming 14 days, including the weather codes (mapped to icons), the maximal and minimal apparent temperatures.

**Offline mode** : as specified above in `WeatherPresenter`, the `CachingStore` holds now a WeatherState that is written to the disk. A weather forecast retrieval is done if the device is online, otherwise it is set by the cached version. This assures the user to have the last retrieved forecast in offline mode.

**Tests** : the `WeatherPresenter` as well as the `OpenMeteoRepository` were tested. The design indeed enabled mocking a `WeatherRepository` and a `WeatherApi`. Tests were added for the cached weather forecast in `CachingStoreTest`. The `WeatherView` was also tested. 

**Error handling** : this branch does not handle weather api error GET responses. For now the WeatherView displays a no forecast icon if there has been some error inside the weather pipeline.


app/src/androidTest/java/com/github/sdpcoachme/database/CachingStoreTest.kt
```diff
@@ -1217,9 +1221,24 @@ class CachingStoreTest {
...
 
-
-
+        val cachingStore = CachingStore(MockDatabase(),
+            ApplicationProvider.getApplicationContext<Context>().dataStoreTest,
+            ApplicationProvider.getApplicationContext()
+        )
+        val target = LatLng(LAUSANNE.latitude, LAUSANNE.longitude)
+        cachingStore.getWeatherForecast(target).thenApply {
+            runBlocking {
```
brygotti, line 1233: Why do we need to wait like this in the callback of the future ?

app/src/main/java/com/github/sdpcoachme/database/CachingStore.kt
```diff
@@ -629,6 +649,29 @@ class CachingStore(private val wrappedDatabase: Database,
...
+     */
+    fun getWeatherForecast(target: LatLng): CompletableFuture<MutableState<WeatherForecast>> {
+
+        return if (isOnline()) {
+            // The WeatherPresenter will launch a weather request and return a future that if
+            // completed normally will update the cache.
+            val weatherPresenter = WeatherPresenter().bind(OpenMeteoRepository(RetrofitClient.api))
+            weatherPresenter.getWeatherForecast(target.latitude, target.longitude).thenApply {
+                weatherForecast = it
+                storeLocalData()
```
brygotti, line 668: by doing a `thenCompose { weatherForecast = it; storeLocalData() }.thenApply { weatherPresenter.observableWeatherForecast }` we can make sure the returned future always completes after the mutablestate has been updated. This could be useful for testing for example.

app/src/main/java/com/github/sdpcoachme/database/CachingStore.kt
```diff
@@ -714,6 +755,9 @@ class CachingStore(private val wrappedDatabase: Database,
         chats.clear()
         cachedTokens.clear()
         contactRowInfos.clear()
+        // we don't clear the weather cache here as getWeatherForecast already makes the
+        // isOnline check and updates the cache accordingly.
+        // todo do the same for the other caches
```
brygotti, line 761: Yes, very good idea.

app/src/main/java/com/github/sdpcoachme/schedule/ScheduleActivity.kt
```diff
@@ -117,10 +99,16 @@ class ScheduleActivity : ComponentActivity() {
 
         val futureDBSchedule: CompletableFuture<Schedule> = store.getSchedule(startMonday)
 
+        // todo move this into caching store
+        val locationProvider = (application as CoachMeApplication).locationProvider
+        locationProvider.updateContext(this, CompletableFuture.completedFuture(null))
+        val userLatLng = locationProvider.getLastLocation().value?: FusedLocationProvider.CAMPUS
+        val weatherState = store.getWeatherForecast(userLatLng).get()
```
brygotti, line 106: The get here is not ideal, although I am not sure how we could remove it. Using CompletableFuture<MutableState>> seems quite hard to deal with actually.

brygotti (approved):
Very nice modularization of the weather pipeline which allows to test individual modules in isolation using mocking. Great job. Some comments have been left regarding testing using asynchronous functions, but given the time constraint, those things will be fixed in the next sprint if it makes sense to do so. Also, as discussed, the weather icon is not present in the GroupEventDetailsActivity, this will also need to be addressed in the next sprint (see #155).

Thanks for the meticulous work.

app/src/androidTest/java/com/github/sdpcoachme/database/CachingStoreTest.kt
Ayskin: As `getWeatherForecast` always returns a `completedFuture`, we need to wait for the observableState to be actually updated.

app/src/main/java/com/github/sdpcoachme/database/CachingStore.kt
Ayskin: Yes, this is way to include the coroutine job completion into a future completion. But then I don't know how to handle the view recomposition on future completion.

app/src/main/java/com/github/sdpcoachme/schedule/ScheduleActivity.kt
Ayskin: Yup, again this follows your other comments, because for now `getWeatherForecast` returns a completedFuture of a mutable weather state.

app/src/androidTest/java/com/github/sdpcoachme/schedule/ScheduleActivityTest.kt
```diff
@@ -382,6 +384,18 @@ class ScheduleActivityTest {
         Intents.release()
     }
 
+    @Test
+    fun weatherForecastIsCorrectlyDisplayedInSchedule() {
+        store.setCurrentEmail(coachEmail).thenApply {
+            ActivityScenario.launch<ScheduleActivity>(defaultIntent).use {
+                composeTestRule.onNodeWithTag("WEATHER_COLUMN")
+                    .assertExists().assertIsDisplayed()
```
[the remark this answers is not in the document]
Ayskin: This changes have been made in upcoming weather-to-event branch

PR #180, opened 2023-05-21
Weather to event

Ayskin:
This small branch includes the weather forecast for a group event page. A test was added in `GroupEventDetailsActivityTest` to test that the view is displayed.

Following the review of the weather-forecast pull request, some not on main thread launched assertions were fixed in tests.

### DIFF
--- a/app/src/main/java/com/github/sdpcoachme/groupevent/GroupEventDetailsActivity.kt
+++ b/app/src/main/java/com/github/sdpcoachme/groupevent/GroupEventDetailsActivity.kt
@@ -48,9 +48,12 @@ import com.github.sdpcoachme.groupevent.GroupEventDetailsActivity.TestTags.Compa
 import com.github.sdpcoachme.groupevent.GroupEventDetailsActivity.TestTags.Companion.TITLE
 import com.github.sdpcoachme.groupevent.GroupEventDetailsActivity.TestTags.Tabs.Companion.ABOUT
 import com.github.sdpcoachme.groupevent.GroupEventDetailsActivity.TestTags.Tabs.Companion.PARTICIPANTS
+import com.github.sdpcoachme.location.provider.FusedLocationProvider
 import com.github.sdpcoachme.messaging.ChatActivity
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/github/sdpcoachme/groupevent/GroupEventDetailsActivity.kt
```diff
@@ -122,6 +125,12 @@ class GroupEventDetailsActivity : ComponentActivity() {
         // do not allow the activity to launch without an id. (double bang)
         val groupEventId = intent.getStringExtra(GROUP_EVENT_ID_KEY)!!
 
+        // todo move this into caching store
+        val locationProvider = (application as CoachMeApplication).locationProvider
+        locationProvider.updateContext(this, CompletableFuture.completedFuture(null))
+        val userLatLng = locationProvider.getLastLocation().value?: FusedLocationProvider.CAMPUS
+        val weatherState = store.getWeatherForecast(userLatLng).get()
```
brygotti, line 132: This could be handled like the other futures and added to the if not null to not have to do a get here. Please correct me if I'm wrong, I'm not sure about, but I think it would work out.

brygotti (changes requested):
Just a small comment to address, and we're all good to merge. Thanks for your work.

app/src/main/java/com/github/sdpcoachme/groupevent/GroupEventDetailsActivity.kt
Ayskin: Yes, I've implemented it this way before. But to be consistent with the `WeatherView` implementation that takes a `MutableState` as a parameter, I would have needed to use `var = remember { ... }` syntax and reassign a new MutableState inside the `LaunchedEffect` instead of changing its value. And as `getWeatherForecast` returns a synchronously completed future, I've decided to leave it as it is. Thanks for the comment and let me know what you think.

brygotti (approved):

PR #181, opened 2023-05-21
Polish dark mode and overall details in the app

brygotti:
This PR mainly consists in adapting all the activities to enable make sure they react correctly to a switch to dark mode in the android system settings. A few things to note:
* Some activities were found to be using a custom TopAppBar "homemade". Those were all converted to using Jetpack Compose' built-in TopAppBar for consistency. This also allows the bar to support dark mode natively. Those activities usually had unexpected behavior when clicking on the back button of the bar (launching an intent to a specific activity instead of just closing the current activity). Those behaviors have been fixed for consistency. Now, clicking on a back button throughout the app always means the same thing: closing the current activity and going back to the previous one.
* The schedule activity now uses the dashboard as a TopAppBar with a custom title, making it more consistent with other main activities of the app. After some testing, I found that keeping the color of the text black for schedule events made the text the most visible. I tried dynamically choosing the color of the text based on the color of the event in the schedule, but having some events with white text and some with black text didn't seem nice to me. The code is still here, commented out if we decide to use it.
* The TopAppBar is now black in dark mode, as opposed to purple, as it is preferred to avoid having large surfaces filled with bright colors in dark mode. This is also the default behavior of the built-in jetpack Compose' TopAppBar.
* The chat activity colors had to be adapted for dark mode. Don't hesitate to give feedback if you don't like them.
* The map colors were also updated to a night theme when in dark mode.
* For the CreateEventActivity, the only issue was about the Dialogs using the maxkeppeler library. Those are using Material 3, which uses a different theme system. I tried enabling material 3 in our app and providing a theme to those components, and everything runs well locally, but some dependency of material 3 seems to need JDK version 9 or above to be compiled, and the CI currently runs JDK 8. A question was posted on Discord for a solution, but for now, the changes regarding material 3 theming were reverted, so that all changes can be merged to main quickly. If necessary, a second PR will be done to fix this, but for now, the maxkeppeler dialogs are displayed in white no matter the theme :(
* The Google Places Autocomplete activity also only partially supports dark mode, and there is nothing that can be done about it.

Other minor changes:
* A good amount of new sports were added to the Sports enum. The SelectSportsList activity was also adapted to better handle a big list of sports. Test also had to be adapted to support scrolling through the sports list that might not always be displayed entirely. A few issues had to be addressed, mainly the fact that the use of LazyColumn in the SelectSportsList activity didn't allow tests to detect compose elements not currently displayed by the list, so it was swapped with a regular Column.
* A loading animation was added to the login activity to make sure the user knows the map activity is being launched.
* Some composables were modularized to used the same design (list item components) in ListItem.kt. This was part of my last PR, but could not be finished properly.
* As per comments in the last PR, documentation was added to those composables and activities related to group events.

Note that the CreateEventActivity looks a bit off compared to the design of the rest of the app. It might need some rework, which was not done here.

### DIFF
--- a/app/src/main/java/com/github/sdpcoachme/database/CachingStore.kt
+++ b/app/src/main/java/com/github/sdpcoachme/database/CachingStore.kt
@@ -731,7 +731,7 @@ class CachingStore(private val wrappedDatabase: Database,
      */
     fun setCurrentEmail(email: String): CompletableFuture<Void> {
         currentEmail = email
-        return completedFuture(null)
+        return storeLocalData()
     }
 
     /**
--- a/app/src/main/java/com/github/sdpcoachme/schedule/CreateEventActivity.kt
+++ b/app/src/main/java/com/github/sdpcoachme/schedule/CreateEventActivity.kt
@@ -1,3 +1,5 @@
+@file:OptIn(ExperimentalMaterial3Api::class)
+
 package com.github.sdpcoachme.schedule
 
 import android.content.Intent
@@ -13,11 +15,11 @@ import androidx.compose.foundation.text.KeyboardActions
 import androidx.compose.foundation.text.KeyboardOptions
 import androidx.compose.material.*
 import androidx.compose.material.icons.Icons
-import androidx.compose.material.icons.filled.Cancel
+import androidx.compose.material.icons.filled.ArrowBack
 import androidx.compose.material.icons.filled.Done
+import androidx.compose.material3.ExperimentalMaterial3Api
 import androidx.compose.runtime.*
 import androidx.compose.ui.Alignment
-import androidx.compose.ui.ExperimentalComposeUiApi
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.toArgb
@@ -40,6 +42,7 @@ import com.github.sdpcoachme.data.schedule.EventType
 import com.github.sdpcoachme.database.CachingStore
 import com.github.sdpcoachme.location.autocomplete.AddressAutocompleteHandler
 import com.github.sdpcoachme.profile.SelectSportsActivity
+import com.github.sdpcoachme.ui.theme.CoachMeMaterial3Theme
 import com.github.sdpcoachme.ui.theme.CoachMeTheme
 import com.maxkeppeker.sheets.core.models.base.Header
 import com.maxkeppeker.sheets.core.models.base.SheetState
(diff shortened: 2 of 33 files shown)


app/src/main/java/com/github/sdpcoachme/database/CachingStore.kt
```diff
@@ -731,7 +731,7 @@ class CachingStore(private val wrappedDatabase: Database,
      */
     fun setCurrentEmail(email: String): CompletableFuture<Void> {
         currentEmail = email
-        return completedFuture(null)
+        return storeLocalData()
```
lindsaybordier, line 734: I think here it would be great to set the email only once retrieving data from the disk has been completed. This would also avoid races, especially in tests if we are not careful, the email set using setCurrentEmail can then be overriden by the email retrieved asynchronously from the disk in another thread.
The code could be as followed:

return retrieveLocalData
    .thenApply { currentEmail = email }
    .thenCompose { storeLocalData() }

app/src/main/java/com/github/sdpcoachme/schedule/CreateEventActivity.kt
```diff
@@ -222,13 +216,14 @@ class CreateEventActivity : ComponentActivity() {
                     },
                     navigationIcon = {
                         IconButton(
-                            onClick = { goBackToScheduleActivity() },
+                            onClick = {
+                                finish() // Go back to previous activity (removes itself from activity stack)
+                            },
```
lindsaybordier, line 221: Great simplification.

lindsaybordier (changes requested):
This is a large PR with a huge amount of work, great effort.
The code is great, not much to say. The overall design of the app has improved and the dark-mode is complete (other than for the createEventActivity for the reason that you mentioned).

Please take a look at my few comments on the code.

I would have a few suggestions for the design (note that they can be done in a next small PR so the code can be merged quickly). We can also discuss those with the team:

- The messages' color (dark blue over black) could be changed. The messages could be black and just have a white or grey delimiter for the contour of the message bubble.
- I think it would be better to keep the weather temperature as integers, as right now too many figures make it hard to read.
- The email could be removed from the dashboard for a more sleek design, and the Dashboard title could be smaller.

Thanks a lot !

app/src/main/java/com/github/sdpcoachme/database/CachingStore.kt
brygotti: A lot of methods in the CachingStore do not follow this pattern (most of them actually). To me, it would make sense for all of them to wait for retrieveLocalData before presenting their result (getUser, updateUser, etc.) to prevent races. Should we update them all ?

lindsaybordier: Yes, we should update them all I agree with that... thanks.
Feel free to write it as an issue if you are not able to take care of this now :)

brygotti: I've added it as an issue #182 so that we can merge this ASAP.

lindsaybordier (approved):

PR #183, opened 2023-05-23
Change the way of redirection of own's coach profile

Thymian18:
# Mini PR
This PR changes the behaviour of clicking on elements that represent one's own coach profile. 
In the case where coaches are in the CoachesListActivity, they can still click on the list element that represents themselves. But now, they get redirected to their own profile which is the same view as if the coach pressed on "My Profile" in the dashboard. So, the coach will be able to see the modifiable view of his own profile instead of just the read-only view. 

The same adaption was made for the case where coaches are in the Map view and click on the infoWindow of their own location.

The diff coverage might not be that good because the code concerning the map cannot be tested because of the known problem of not being able to test markers.


Luca-Engel (approved):
This is a small pull request. Nevertheless, it increases the flow of our program a lot as a coach still wants to see if they are displayed on the map/in the nearby coaches section.
Also, this pr prevents chatcreations with yourself which caused problems before.

Great job!

PR #184, opened 2023-05-23
New contact messaging contacts bugfix

Luca-Engel:


app/src/androidTest/java/com/github/sdpcoachme/database/CachingStoreTest.kt
brygotti, line 1: As discussed already, tests that run asynchronously like this are very hard to debug without using a lot println's... It would make more sense to use .get() with a timeout to see more easily where the test fails without having to re-run it and add println's. However, given that all of them are written like this in this test class, I understand that you might not have enough time to change them all. Thus, it might make more sense to create an issue for the matter and merge the PR quickly. Do as you prefer!

app/src/main/java/com/github/sdpcoachme/database/CachingStore.kt
```diff
@@ -507,6 +506,30 @@ class CachingStore(private val wrappedDatabase: Database,
...
     }
 
+    /**
+     * Updates the chat contacts of the current user if the given contact is not already in the list
+     *
+     * @param email The email of the current user
+     * @param chatId The potentially new contact to add
+     */
+    fun addChatContactIfNew(email: String, chatId: String, contact: String): CompletableFuture<Void> {
+        println("Adding chat contact if new called on $email with chatid $chatId and contact $contact")
```
brygotti, line 516: println's can be removed in this method

app/src/main/java/com/github/sdpcoachme/database/FireDatabase.kt
```diff
@@ -90,9 +90,12 @@ class FireDatabase(databaseReference: DatabaseReference) : Database {
     }
 
     override fun getContactRowInfos(email: String): CompletableFuture<List<ContactRowInfo>> {
+        println("getContactRowInfos email: $email")
```
brygotti, line 93: println's in this method can be removed too

brygotti (changes requested):
Nice work for this fix. Please have a look at the few comments (mostly details), and then we can merge. Thanks!

app/src/androidTest/java/com/github/sdpcoachme/database/CachingStoreTest.kt
Luca-Engel: I have also thought about this idea but I believe that it is much more natural to not include gets on every line that uses futures as this limits the possibilities of testing a lot.
Say, you wanted to test how the app functions when there are delays in these operations (just like in one of the SwEng exams). This is only possible if we do not include get operations in every test. Therefore, I believe it is more modular to leave out the get calls in the tests.

brygotti (approved):

PR #185, opened 2023-05-24
Push notification bugfix

Luca-Engel:
This pull request addresses the bug where the application would crash when the user logged out after receiving a push notification when inside the app and then clicked on it after having logged out. This crash happened as there was no user email in the application. This has been fixed now.

Also, the redirection of the user when clicking on the push notification has been adapted to route the user through the contacts listing activity to ensure that the app would not close on the back button press when freshly opening the application by clicking on the push notification. Now, when users click on the notification and then the back button, they will be redirected to the contacts listing.

The added changes have also been tested.

### DIFF
--- a/app/src/androidTest/java/com/github/sdpcoachme/messaging/ChatActivityTest.kt
+++ b/app/src/androidTest/java/com/github/sdpcoachme/messaging/ChatActivityTest.kt
@@ -441,6 +441,23 @@ class ChatActivityTest {
         }
     }
 
+    // simulates push notifications where the user logs out after receiving the notification and then logs in
+    @Test
+    fun chatActivityRecoversFromMissingCurrentUserEmailIfItIsPassedInTheIntentExtra() {
+        store.setCurrentEmail("").get(1000, TimeUnit.MILLISECONDS)
+        val emailRecoveryIntent = personalChatDefaultIntent
+            .putExtra("pushNotification_currentUserEmail", currentUser.email)
+
+        ActivityScenario.launch<ChatActivity>(emailRecoveryIntent).use {
+            waitForLoading(it)
+
+            composeTestRule.onNodeWithTag(BACK).assertIsDisplayed()
+            composeTestRule.onNodeWithTag(CONTACT_FIELD.LABEL, useUnmergedTree = true).assertTextEquals(toUser.firstName + " " + toUser.lastName)
+            composeTestRule.onNodeWithTag(CHAT_FIELD.LABEL, useUnmergedTree = true).assertIsDisplayed()
+            composeTestRule.onNodeWithTag(CHAT_BOX.CONTAINER).assertIsDisplayed()
+        }
+    }
+
 
     // Waits for the activity to finish loading any async state
     private fun waitForLoading(scenario: ActivityScenario<ChatActivity>) {
--- a/app/src/androidTest/java/com/github/sdpcoachme/messaging/InAppNotifierTest.kt
+++ b/app/src/androidTest/java/com/github/sdpcoachme/messaging/InAppNotifierTest.kt
@@ -148,6 +148,11 @@ class InAppNotifierTest {
         }
     }
 
+    // This test checks that the emails of the users can be recovered when receiving a push notification while the app is
+    // open, then logging out and closing the application. After closing the application, the notification is pressed.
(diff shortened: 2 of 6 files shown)


app/src/androidTest/java/com/github/sdpcoachme/messaging/ChatActivityTest.kt
```diff
@@ -441,6 +441,23 @@ class ChatActivityTest {
         }
     }
 
+    // simulates push notifications where the user logs out after receiving the notification and then logs in
```
Thymian18, line 444: Very good to have that comment for the understanding. Otherwise, it would be a bit hard to completely understand what behaviour is tested here.

app/src/androidTest/java/com/github/sdpcoachme/messaging/InAppNotifierTest.kt
```diff
@@ -145,6 +148,75 @@ class InAppNotifierTest {
         }
     }
 
+    @Test
+    fun onMessageReceivedEnablesEmailRecoveryAndOpensChatActivityWhenUserReceivesNotificationThenLoggsOutAndThenClicksOnIt() {
```
Thymian18, line 152: Regarding the long test name, it would be good to have a small comment that describes what the test is for and maybe why it is necessary.

app/src/androidTest/java/com/github/sdpcoachme/messaging/InAppNotifierTest.kt
```diff
@@ -145,6 +148,75 @@ class InAppNotifierTest {
...
+
+            // Check if ChatActivity is opened
+            composeTestRule.onNodeWithText(toUser.firstName + " " + toUser.lastName).assertExists()
+            composeTestRule.onNodeWithTag(CONTACT_FIELD.LABEL, useUnmergedTree = true).assertExists()
+            composeTestRule.onNodeWithTag(CHAT_FIELD.LABEL, useUnmergedTree = true).assertExists()
+        }
+    }
+
+    @Test
+    fun onMessageReceivedEnablesEmailRecoveryAndOpensContactsListWhenUserReceivesNotificationWithoutChatIdThenLoggsOutAndThenClicksOnIt() {
```
Thymian18, line 187: Good that you thought about this case too. I think, once the above test has a well-explanatory comment, this test will be understandable too.

app/src/main/java/com/github/sdpcoachme/messaging/ChatActivity.kt
```diff
@@ -109,6 +109,13 @@ class ChatActivity : ComponentActivity() {
 
         store = (application as CoachMeApplication).store
         emailFuture = store.getCurrentEmail()
+            .exceptionally {
```
Thymian18, line 112: Yes, this is the code we need for the bugfix! Also, the comment makes it clear what exactly the bug was about and how it is fixed.

app/src/main/java/com/github/sdpcoachme/messaging/InAppNotifier.kt
```diff
@@ -59,18 +60,25 @@ class InAppNotifier(val context: Context, val store: CachingStore) {
     private fun sendMessagingNotification(notificationTitle: String, notificationBody: String, chatId: String) {
 
         // The more info we receive, the more we can customize the notification's behaviour (up until the chat itself)
-        store.isLoggedIn().thenAccept { isLoggedIn ->
+        store.isLoggedIn().thenCompose { isLoggedIn ->
+            if (isLoggedIn) store.getCurrentEmail() else CompletableFuture.completedFuture("")
+        }.thenAccept { email ->
             val intent = when {
-                !isLoggedIn ->
+                email.isEmpty() ->
```
Thymian18, line 67: In the case where email is empty, what happens if the chatId is empty too? In the case where email is not empty, we have that if check but for the case of the empty email, we don't have it. Maybe you could check if that check is made internally or if you have to add an if check here.

Luca-Engel: This check is already conducten in the LoginActivity and handled accordingly

Thymian18 (changes requested):
There are only very few minor adaptions to make, other than that it seems clean and fixes a quite important bug.

Thymian18 (approved):
LGTM

PR #186, opened 2023-05-25
Refine design

Thymian18:
In this PR, I worked on three things: the design of the dashboard, the design of the createEvent form, and the position of the add event drop down. 

## Dashboard Design
The new dashboard displays the name as well as the email and profile picture of the user. Additionally, the colors were adapted according to light/dark mode. 

image
image

## CreateEvent Form Design
The form now has the same design elements as used in ProfileActivity (attribute rows). As the elements are not the same composables, I needed to adapt quite some tests and spent a lot of time on managing all the test tags and labels. Also, the form is now scrollable so that in case of a long description, the user can still see the whole description. 

image

## Add Event Dropdown Position
The dropdown menu is now positioned just above the '+' floating button as shown in the following image. 

image

## Comment
One test in ScheduleActivityTest is not really good, but I'll fix it in another PR (if I have enough time) where I am also gonna adapt the position of the DropDown to choose an event type.


app/src/androidTest/java/com/github/sdpcoachme/schedule/ScheduleActivityTest.kt
```diff
@@ -82,20 +82,21 @@ class ScheduleActivityTest {
         }
     }
 
-    @Test
+    // TODO: Fix this test
```
Luca-Engel, line 85: It might be helpful to add a comment to this test to explain why it is commented out now.

app/src/main/java/com/github/sdpcoachme/ui/Dashboard.kt
```diff
@@ -279,35 +326,87 @@ fun AppBar(title: @Composable (Modifier) -> Unit, onNavigationIconClick: () -> U
     )
 }
 
+/**
+ * Composable that displays the header of the drawer.
+ * It contains the user's profile picture, name and email.
```
Luca-Engel, line 331: Great idea to add javadocs here, this makes it much more clear what to expect from this composable!

Luca-Engel (commented):
Great job on this pull request! The refactoring you did really enhances the quality and feel of our application a lot!
There is just a little question I had about clarifying a commented out test.
Other than that, it looks great, nice job!

app/src/androidTest/java/com/github/sdpcoachme/schedule/ScheduleActivityTest.kt
Thymian18: Yes, this test is currently commented out because it introduces problems for other tests. The time was too short to fix this, but considering the nondeterministic behaviour of all other tests, it might be the best to leave it out for now. Also, the coverage does not suffer much from this.

Luca-Engel (approved):
Thanks for the clarification! Ready for merge :)

PR #188, opened 2023-05-25
Rating backend

lindsaybordier:
This Pull Request consists of the backend code for the rating functionality, as well as some of the frontend as time permitted, to allow @Ayskin to focus more on certain parts of the rating frontend.

Important notes on the rating functionality:

- For every coach, we keep track of all the ratings given by users.
- Every rating given by a user to a coach is an integer between 0 and 5 included.
- The rating of a coach is the average of the rating given to them rounded to the nearest integer.
- If a user gives a new rating to a coach they already rated, it overrides the previous rating for this user (only one rating per user is kept).
- The functionality is implemented to work even if the coach has no rating field yet in the database and work properly in offline mode.

Concerning the frontend, I added the rating stars label to the Nearby Coaches (CoachesListActivity) and to the coach bubbles appearing when we click on their localisation markers on the main map (MapActivity).


app/src/androidTest/java/com/github/sdpcoachme/database/CachingStoreTest.kt
```diff
@@ -79,6 +81,105 @@ class CachingStoreTest {
...
+            .get(1, SECONDS)
+        assertThat(averageRating3, `is`(3))
+    }
+
+    @Test
+    fun addRatingToNonCoachUserThrowsException() {
+        cachingStore.setCurrentEmail(NON_COACH_1.email).get(1, SECONDS)
+        cachingStore.updateUser(NON_COACH_2).get(1, SECONDS)
+        try {
+            cachingStore.addRatingToCoach(NON_COACH_2.email, 5).get(1, SECONDS)
```
Luca-Engel, line 143: Just as a general comment, I believe that using assertThrows could be written more concise than the assertion done here with the try and catch block. At the end of the day, your code does the same thing, just a little less concise.

app/src/main/java/com/github/sdpcoachme/data/Sports.kt
```diff
@@ -24,4 +24,5 @@ enum class Sports(val sportName: String, val sportIcon: ImageVector) {
     VOLLEYBALL("Volleyball", Icons.Default.SportsVolleyball),
     WORKOUT("Workout", Icons.Default.FitnessCenter),
     YOGA("Yoga", Icons.Default.SelfImprovement),
+    WATERSKI("Waterski", Icons.Default.DownhillSkiing),
```
Luca-Engel, line 27: Gotta represent ;)

Luca-Engel (approved):
Very nicely done! This feature is super cool and agments the user experience a lot!

I have added a little comment on how you test what exceptions are thrown and how it could be rewritten a little more consisely, however, as this is a minor detail and does not decrease the readability of the tests too much, this is no problem.

Well done!

app/src/androidTest/java/com/github/sdpcoachme/database/CachingStoreTest.kt
lindsaybordier: Indeed you are totally right, however assertThrows is only available in JUNIT 5 and we are using JUNIT 4.

I spent a quiet a lot of time figuring out how to do these tests without assertThrows and this is the cleanest solution I could come up with.

I did not change the JUNIT version to avoid creating problems with the testing suite at the last moment... But this can definitely be done later and a small refactor of those tests will make them way more concise.

PR #189, opened 2023-05-25
Add local profile pictures to users and group chats, and a few bugfixes

brygotti:
### Profile pictures
This PR adds support for local profile pictures. They are stored as resources in the drawable folder, and for each user (group event), a profile picture is chosen based on the hash of the user's email (group event's id) modulo the number of pictures available. As of now, there are 8 pictures available for the profiles (resources with a name that starts with profile_picture_), and they consist of fishes (😄), while there are 4 pictures available for group events (displayed for the associated group chat, resources with a name that starts with group_event_picture), which consist of cats (😻). We can always add/change the profile pictures used in a future PR, to adapt them for the demo). Methods were added to UserInfo and GroupEvent to get the resource id of the corresponding picture. If we ever decide to change this system with real profile pictures, we will need to change this code, that is why TODOs were left to indicate it as temporary.
Given the structure of ContactRowInfo (which does not contain participants of a Chat) it was hard to adapt the code to retrieve the correct picture in case of non group chat conversations. Here too, TODOs were left to indicate the code as temporary.
For UserInfo that have an empty email or GroupEvent that have an empty id, the picture is not chosen as the hash but rather as a hardcoded gray image. This allows to handle cases where in certain composables, an empty UserInfo or an empty GroupEvent is used as placeholder state before the state loads. In this case, we display a gray image before the state loading is done.

### Bug fixes and polishing of the app
* Fix padding issues with status tick (read, received, etc.) in messages displayed in ChatActivity
* Fix overflow issues with long text, all over the app
* Fix overflow issues with long list of sports, all over the app
* Fix bug where activities would not update their state on resuming to the foreground (now creating a new event instantly shows it in the schedule without having to reload the schedule activity, and joining a group event is instantly reflected in GroupEventsList activity.)
* Fix wrong behavior of back button from activities with Dashboard


app/src/main/java/com/github/sdpcoachme/database/CachingStore.kt
```diff
@@ -487,58 +469,56 @@ class CachingStore(private val wrappedDatabase: Database,
...
-                                    start >= minCachedMonday && end <= maxCachedMonday
-                                } else {
-                                    false
-                                }
+        return getCurrentEmail().thenCompose { email ->
+            if (cachedSchedule.events.isEmpty() && cachedSchedule.groupEvents.isEmpty()) {  // If no cached schedule for that account, we fetch the schedule from the db
+                wrappedDatabase.getSchedule(email).thenApply { schedule ->
+                    val events =
+                        schedule.events.filter {   // We only cache the events that are in the current week or close to it
+                            if (it != Event()) {
```
Luca-Engel, line 477: It would be nice to refactor the code to remove code block duplication but, due to time constraints, I believe it is best to move this to a later time and merge this pr.

Luca-Engel (approved):
Great job on this pull request! You added a really nice feature and did quite a bit of minor changes to the code.  
I added a comment about code climate issues saying that it would be best to fix these issues, however, due to time constraints, we can move this to a later time.

PR #190, opened 2023-05-26
Rating frontend

Ayskin:
**Feature** : this PR adds the displaying of a coach average rating on its profile. The retrieval  and update of the rating is done thanks to the backend rating part written by @lindsaybordier.

**Implementation** : the `RatingBar` displays 5 stars in a coach `ProfileActivity` and can be changed via a `RatingActivity`. The design follows the same structure as for `SelectSportsActivity` and `EditTextActivity`.

**Testing** : tests were added to test the `RatingView` that wraps the `RatingBar` as well as for the behavior of the `RatingActivity`.

**Quickfix** : lindsay helped me to add a quickfix where the database ratings fields were not correctly set in firebase.


app/src/main/java/com/github/sdpcoachme/profile/ProfileActivity.kt
```diff
@@ -138,12 +156,21 @@ class ProfileActivity : ComponentActivity() {
      */
     @Composable
     fun Profile(futureUserInfo: CompletableFuture<UserInfo>, isViewingCoach: Boolean) {
+
         val context = LocalContext.current
         var userInfo by remember { mutableStateOf(UserInfo()) }
+        var coachRating by remember { mutableStateOf(0)}
+        var currIsCoach by remember { mutableStateOf(false) }
```
brygotti, line 163: could be using the userInfo.coach variable instead of redefining another variable currIsCoach

app/src/main/AndroidManifest.xml
```diff
@@ -24,6 +24,11 @@
         android:supportsRtl="true"
         android:theme="@style/Theme.CoachMe"
         tools:targetApi="31">
+        <activity
+            android:name=".rating.RatingActivity"
+            android:exported="false"
+            android:label="@string/title_activity_group_events_list"
```
brygotti, line 30: Probably need to change the title here to something more appropriate

app/src/main/java/com/github/sdpcoachme/profile/ProfileActivity.kt
```diff
@@ -194,6 +221,31 @@ class ProfileActivity : ComponentActivity() {
         ) {
             TitleRow(userInfo, isViewingCoach)
             Spacer(modifier = Modifier.height(10.dp))
+            if (currIsCoach || isViewingCoach) {
```
brygotti, line 224: with this implementation, it shows a rating even for non-coaches that are displayed via the ProfileActivity with isViewingCoach set to true (the default behavior through out the app). Would probably make more sense to only display the rating when currIsCoach (or userInfo.coach) without taking into account whether isViewingCoach is true or not.

app/src/main/java/com/github/sdpcoachme/rating/RatingActivity.kt
brygotti, line 1: All those "edit" activities are very similar, code should be merged into one at some point (might want to add a TODO for that)

app/src/main/java/com/github/sdpcoachme/rating/RatingView.kt
```diff
@@ -0,0 +1,147 @@
...
+@Composable
+fun RatingView(
+    title: String,
+    initialRating: Int,
+    onCancel: () -> Unit,
+    onSubmit: (Int) -> Unit
+) {
+    var rating by remember { mutableStateOf(initialRating) }
+    var selected by remember { mutableStateOf(false) }
+    val size by animateDpAsState(
```
brygotti, line 58: Gotta love the animation! 😃

brygotti (approved):
A few comments are left and could be addressed in a future PR, but given time constraints, we need to merge this ASAP. Overall great work, thanks!

